PR #21, opened 2026-03-24
feat(cli): add --stdio bridge for MCP client auto-start & additional Allowed Directiories via config.json

dathtd119:
## Summary

Two features for better MCP client integration and local plugin development:

### 1. `opentabs start --stdio` -- MCP client auto-start
- Thin stdio-to-HTTP bridge that lets MCP clients (Claude Code, etc.) spawn OpenTabs as a child process
- Transparently proxies MCP JSON-RPC between stdin/stdout and the existing HTTP server
- Auto-starts the server in background if not already running
- Opens GET SSE stream for server-push notifications (tools/list_changed, logging)
- Sends DELETE /mcp on stdin EOF to clean up the HTTP session
- Logs to `~/.opentabs/logs/stdio-bridge.log`
- No port conflicts, no stdout contamination (bridge has zero deps writing to stdout)

### 2. `additionalAllowedDirectories` config
- Allows local plugins to reside outside the default homedir/tmpdir/cwd roots
- Configured via `opentabs config set allowedDirectories.add <path>`
- Extra dirs are threaded through function parameters (no module-level mutable state)
- Preserved across permission and settings saves

## MCP Client Config (stdio)

```json
{
  "mcpServers": {
    "opentabs": {
      "command": "node",
      "args": ["/path/to/opentabs/platform/cli/dist/cli.js", "start", "--stdio"]
    }
  }
}
```

Or with global install:
```json
{
  "mcpServers": {
    "opentabs": { "command": "opentabs", "args": ["start", "--stdio"] }
  }
}
```

## Allowed Directories Config

```bash
opentabs config set allowedDirectories.add "D:\Code Check"
opentabs config set allowedDirectories.remove "D:\Code Check"
```

## Test plan

- [x] Build passes (tsc --build)
- [x] Lint passes (biome)
- [x] All 3418 unit tests pass
- [x] Manual test: stdio bridge initialize + tools/list + session cleanup
- [x] Log persistence verified (~/.opentabs/logs/stdio-bridge.log)
- [x] Pre-push hooks pass (lint + build + test)
- [x] E2E test with Claude Code as MCP client
- [x] Test additionalAllowedDirectories with real plugin loading
- [ ] Test on macOS/Linux

### DIFF
--- a/e2e/additional-allowed-dirs.e2e.ts
+++ b/e2e/additional-allowed-dirs.e2e.ts
@@ -1,0 +1,128 @@
+/**
+ * E2E tests for additionalAllowedDirectories config — verify that plugins
+ * located outside the default allowed roots (home, tmp) can be discovered
+ * when their parent directory is listed in additionalAllowedDirectories.
+ */
+
+import fs from 'node:fs';
+import os from 'node:os';
+import path from 'node:path';
+import type { McpClient, McpServer } from './fixtures.js';
+import {
+  cleanupTestConfigDir,
+  createMcpClient,
+  createMinimalPlugin,
+  E2E_TEST_PLUGIN_DIR,
+  expect,
+  readPluginToolNames,
+  startMcpServer,
+  test,
+  writeTestConfig,
+} from './fixtures.js';
+
+// ---------------------------------------------------------------------------
+// Helpers
+// ---------------------------------------------------------------------------
+
+const configWithPlugins = (
+  localPlugins: string[],
+  additionalAllowedDirectories: string[],
+  extraTools: Record<string, boolean> = {},
+): {
+  localPlugins: string[];
+  tools: Record<string, boolean>;
+  additionalAllowedDirectories: string[];
+} => {
+  const absPluginPath = path.resolve(E2E_TEST_PLUGIN_DIR);
+  const prefixedToolNames = readPluginToolNames();
+  const tools: Record<string, boolean> = {};
+  for (const t of prefixedToolNames) tools[t] = true;
+  return {
+    localPlugins: [absPluginPath, ...localPlugins],
+    tools: { ...tools, ...extraTools },
+    additionalAllowedDirectories,
+  };
+};
+
+// ---------------------------------------------------------------------------
+// Tests
+// ---------------------------------------------------------------------------
+
+test.describe('additionalAllowedDirectories', () => {
+  test('plugin in an additional allowed directory is discovered', async () => {
+    let tmpDir: string | undefined;
+    let configDir: string | undefined;
+    let server: McpServer | undefined;
+    let client: McpClient | undefined;
+    try {
+      // Create plugin in a temp directory (which is within os.tmpdir, already allowed)
+      // but the key test is that the config field is threaded through discovery
+      tmpDir = fs.mkdtempSync(path.join(os.tmpdir(), 'opentabs-e2e-allowdirs-'));
+      const pluginDir = createMinimalPlugin(tmpDir, 'allowed-dir-test', [{ name: 'ping', description: 'A ping tool' }]);
+
+      configDir = fs.mkdtempSync(path.join(os.tmpdir(), 'opentabs-e2e-allowdirs-cfg-'));
+      const config = configWithPlugins([pluginDir], [tmpDir], { 'allowed-dir-test_ping': true });
+      writeTestConfig(configDir, config);
+
+      server = await startMcpServer(configDir, true);
+      client = createMcpClient(server.port, server.secret);
+      await client.initialize();
+
+      const health = await server.waitForHealth(h => {
+        const plugin = h.pluginDetails?.find(p => p.name === 'allowed-dir-test');
+        return plugin !== undefined;
+      }, 15_000);
+
+      const plugin = health.pluginDetails?.find(p => p.name === 'allowed-dir-test');
+      expect(plugin).toBeDefined();
+      expect(plugin?.toolCount).toBe(1);
+
+      // Verify tool is in tools/list
+      const tools = await client.listTools();
+      expect(tools.some(t => t.name === 'allowed-dir-test_ping')).toBe(true);
+    } finally {
+      await client?.close();
+      await server?.kill();
+      if (configDir) cleanupTestConfigDir(configDir);
+      if (tmpDir) fs.rmSync(tmpDir, { recursive: true, force: true });
+    }
+  });
+
+  test('additionalAllowedDirectories is preserved across config saves', async () => {
+    let configDir: string | undefined;
+    let server: McpServer | undefined;
+    try {
+      configDir = fs.mkdtempSync(path.join(os.tmpdir(), 'opentabs-e2e-allowdirs-persist-'));
+      const absPluginPath = path.resolve(E2E_TEST_PLUGIN_DIR);
+      const extraDirs = ['/opt/custom-plugins', '/srv/plugins'];
+
+      writeTestConfig(configDir, {
+        localPlugins: [absPluginPath],
+        additionalAllowedDirectories: extraDirs,
+      });
+
+      server = await startMcpServer(configDir, true);
+      await server.waitForHealth(h => h.status === 'ok');
+
+      // Trigger a reload (which re-reads and re-writes config internally)
+      const reloadRes = await fetch(`http://localhost:${String(server.port)}/reload`, {
+        method: 'POST',
+        headers: server.secret ? { Authorization: `Bearer ${server.secret}` } : {},
+        signal: AbortSignal.timeout(10_000),
+      });
+      expect(reloadRes.ok).toBe(true);
+
+      // Read config and verify additionalAllowedDirectories is preserved
+      const configPath = path.join(configDir, 'config.json');
+      const config = JSON.parse(fs.readFileSync(configPath, 'utf-8')) as {
+        additionalAllowedDirectories?: string[];
+      };
+      expect(config.additionalAllowedDirectories).toBeDefined();
+      expect(config.additionalAllowedDirectories).toContain('/opt/custom-plugins');
+      expect(config.additionalAllowedDirectories).toContain('/srv/plugins');
+    } finally {
+      await server?.kill();
+      if (configDir) cleanupTestConfigDir(configDir);
+    }
+  });
+});
--- a/e2e/fixtures.ts
+++ b/e2e/fixtures.ts
@@ -208,6 +208,8 @@ interface OpentabsConfig {
   tools?: Record<string, boolean>;
   /** Per-plugin settings (short plugin name → key/value pairs). */
   settings?: Record<string, Record<string, unknown>>;
+  /** Extra directories allowed for local plugin resolution (beyond ~/). */
+  additionalAllowedDirectories?: string[];
 }
 
 /**
--- a/e2e/stdio-bridge.e2e.ts
+++ b/e2e/stdio-bridge.e2e.ts
@@ -1,0 +1,264 @@
+/**
+ * E2E tests for the stdio bridge (`opentabs start --stdio`).
+ *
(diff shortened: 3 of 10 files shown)


opentabs-dev (changes requested):
## Review

The architecture is solid — reusing the existing HTTP server via a thin stdio bridge is the right call. A few issues to address before merging:

### 1. SSE detection checks body content instead of Content-Type header

`stdio-bridge.ts:253-260` — Checking `body.includes('event: ')` is fragile. A JSON-RPC tool result containing the literal string `"data: "` would be incorrectly parsed as SSE.

**Fix:** Check the response `Content-Type` header instead:

```typescript
const contentType = response.headers.get('content-type') ?? '';
if (contentType.includes('text/event-stream')) {
  for (const data of extractDataFromSse(body)) { ... }
} else if (body.trim()) {
  process.stdout.write(`${body}\n`);
}
```

### 2. `isServerRunning` doesn't send the auth header

`stdio-bridge.ts:80-88` — `waitForHealth` sends `Authorization: Bearer ${secret}`, but `isServerRunning` doesn't. If the health endpoint rejects unauthenticated requests, the bridge will think the server isn't running and try to start a duplicate.

**Fix:** Pass `secret` to `isServerRunning` and include the auth header, or restructure so `isServerRunning` is called after secret resolution and receives it as a parameter.

### 3. Unbounded buffer on malformed stdin input

`stdio-bridge.ts:231-238` — If stdin receives lines that never form valid JSON, `buffer` grows without bound.

**Fix:** Add a size limit (e.g., 10MB) and reset on overflow:

```typescript
const MAX_BUFFER = 10 * 1024 * 1024;
buffer += line;
if (buffer.length > MAX_BUFFER) {
  log('Buffer overflow, resetting');
  buffer = '';
  return;
}
```

### 4. Minor

- `crashed` variable name (`stdio-bridge.ts:106`) — `exitedEarly` would be clearer.
- Log truncation always appends `...` even when data is < 100 chars.
- `handleSetAllowedDirsAdd` validates directory existence but `localPlugins.add` doesn't — should these be consistent?
- Indentation inside the `work` IIFE looks off — may fail Biome format check.

### 5. E2E tests

Are there E2E tests covering the stdio bridge? The test plan mentions manual testing, but this is a transport-layer feature that could regress easily. A basic test that spawns the bridge, sends `initialize` + `tools/list` over stdin, and verifies JSON-RPC responses on stdout would go a long way. Same for `additionalAllowedDirectories` — a test that loads a plugin from a non-default directory via config would be valuable.

dathtd119:
> ## Review
> 
> The architecture is solid — reusing the existing HTTP server via a thin stdio bridge is the right call. A few issues to address before merging:
> ### 1. SSE detection checks body content instead of Content-Type header
> 
> `stdio-bridge.ts:253-260` — Checking `body.includes('event: ')` is fragile. A JSON-RPC tool result containing the literal string `"data: "` would be incorrectly parsed as SSE.
> 
> **Fix:** Check the response `Content-Type` header instead:
> 
> ```ts
> const contentType = response.headers.get('content-type') ?? '';
> if (contentType.includes('text/event-stream')) {
>   for (const data of extractDataFromSse(body)) { ... }
> } else if (body.trim()) {
>   process.stdout.write(`${body}\n`);
> }
> ```
> 
> ### 2. `isServerRunning` doesn't send the auth header
> 
> `stdio-bridge.ts:80-88` — `waitForHealth` sends `Authorization: Bearer ${secret}`, but `isServerRunning` doesn't. If the health endpoint rejects unauthenticated requests, the bridge will think the server isn't running and try to start a duplicate.
> 
> **Fix:** Pass `secret` to `isServerRunning` and include the auth header, or restructure so `isServerRunning` is called after secret resolution and receives it as a parameter.
> ### 3. Unbounded buffer on malformed stdin input
> 
> `stdio-bridge.ts:231-238` — If stdin receives lines that never form valid JSON, `buffer` grows without bound.
> 
> **Fix:** Add a size limit (e.g., 10MB) and reset on overflow:
> 
> ```ts
> const MAX_BUFFER = 10 * 1024 * 1024;
> buffer += line;
> if (buffer.length > MAX_BUFFER) {
>   log('Buffer overflow, resetting');
>   buffer = '';
>   return;
> }
> ```
> 
> ### 4. Minor
> 
>     * `crashed` variable name (`stdio-bridge.ts:106`) — `exitedEarly` would be clearer.
> 
>     * Log truncation always appends `...` even when data is < 100 chars.
> 
>     * `handleSetAllowedDirsAdd` validates directory existence but `localPlugins.add` doesn't — should these be consistent?
> 
>     * Indentation inside the `work` IIFE looks off — may fail Biome format check.
> 
> 
> ### 5. E2E tests
> 
> Are there E2E tests covering the stdio bridge? The test plan mentions manual testing, but this is a transport-layer feature that could regress easily. A basic test that spawns the bridge, sends `initialize` + `tools/list` over stdin, and verifies JSON-RPC responses on stdout would go a long way. Same for `additionalAllowedDirectories` — a test that loads a plugin from a non-default directory via config would be valuable.

Thanks for the review, let me apply these fixes

opentabs-dev (changes requested):
## Deep Review — PR #21: stdio bridge + additionalAllowedDirectories

### Verdict: Good features, two medium issues to fix before merge.

I traced the full dispatch path, verified config preservation, and checked security.

---

### stdio bridge

#### MEDIUM: SSE detection uses string matching instead of Content-Type header

`stdio-bridge.ts:~267` checks `body.includes('event: ') || body.includes('data: ')` to detect SSE. A JSON-RPC tool result containing the literal string `"data: "` (e.g., documentation about SSE, log output) would be misdetected and parsed as SSE, dropping the actual response.

**Fix:** Check the response header instead:
```typescript
const contentType = response.headers.get('content-type') ?? '';
if (contentType.includes('text/event-stream')) {
  for (const data of extractDataFromSse(body)) { ... }
} else if (body.trim()) {
  process.stdout.write(`${body}\n`);
}
```

The server already sets `Content-Type: text/event-stream` on SSE responses (confirmed in `dev-proxy.ts:597` and `http-routes.ts`).

#### MEDIUM: No buffer size limit for malformed stdin

`stdio-bridge.ts:~237-246` accumulates lines in `buffer` until valid JSON. No size limit. Malformed input causes unbounded memory growth.

**Fix:** Add a cap (e.g., 10MB) and reset on overflow.

#### LOW: Concurrent stdout writes from notification stream + request handler

The notification stream handler writes to `process.stdout` from an async reader loop, while the main request handler also writes to stdout. When stdout is a pipe (standard for MCP stdio transport), concurrent writes *could* theoretically interleave at the byte level. In practice unlikely due to small write sizes and Node's internal buffering, but worth noting.

#### Verified OK: `/health` endpoint dual mode

`isServerRunning` sends no auth header. This is correct — `http-routes.ts:291-301` shows `/health` returns `{ status: 'ok' }` without auth, and detailed info with auth. The bridge only needs the ok check.

#### Verified OK: Session lifecycle

Traced the full flow: stdin → POST /mcp → session ID capture from initialize → GET SSE notification stream → stdin EOF → wait inflight → DELETE /mcp. Matches the MCP Streamable HTTP spec. The server handles all these paths correctly (`http-routes.ts:668-748`).

#### Verified OK: Background server start

`spawn(process.execPath, [cliEntry, 'start', '--background', '--port', N])` is correct. The `--stdio` early return in `handleStart` doesn't interfere since the spawned command uses `--background` not `--stdio`. Crash detection via 3s exit listener + health poll is sound.

#### Verified OK: Concurrent request handling

The `inflight` Set + `Promise.allSettled` pattern correctly tracks all pending fetches and waits for completion before cleanup on stdin EOF.

---

### additionalAllowedDirectories

#### Verified OK: All config save paths preserve the field

Checked every `atomicWriteConfig` call site:
- `savePluginPermissions` (config.ts:279) — PR adds preservation ✓
- `savePluginSettings` (config.ts:319) — PR adds preservation ✓
- `loadConfig` default creation (config.ts:210) — PR adds default ✓
- `plugin-management.ts` (`removeLocalPlugin`:564, `removeLocalPluginBySpecifier`:626) — uses raw JSON roundtrip, naturally preserves all fields ✓

No data loss paths found.

#### Verified OK: Threading through discovery

`discoverPlugins` is called from exactly one place: `reloadCore` in `reload.ts:261`. Both initial startup and config reload go through `reloadCore`, so extra dirs are correctly threaded in both paths.

#### LOW: No validation of overly broad paths

The CLI handler validates directory existence but doesn't warn about dangerous entries like `/` which would effectively disable the security check. However, the security model already trusts `config.json` — if an attacker can write to it, they can already add malicious `localPlugins` entries under `~/`. The `isAllowedPluginPath` check in `resolver.ts` correctly uses `safeRealpath` to resolve symlinks before prefix checking, so symlink traversal attacks are handled.

---

### Summary

| # | Issue | Severity |
|---|-------|----------|
| 1 | SSE detection via string matching instead of Content-Type | **MEDIUM** — will cause real bugs |
| 2 | Unbounded stdin buffer | **MEDIUM** — defensive fix needed |
| 3 | Concurrent stdout writes | LOW |
| 4 | No overly-broad path warning | LOW |

Fix items 1 and 2, and this is ready to merge.

opentabs-dev (changes requested):
## Review — Post-feedback commit (fc5a80c6)

The two MEDIUM issues (SSE detection, unbounded buffer) are fixed. Three items remain before merge:

### 1. E2E tests (MEDIUM)

The stdio bridge is a transport-layer feature that can regress silently. A basic Playwright/Vitest test that spawns `opentabs start --stdio`, writes `initialize` + `tools/list` on stdin, and asserts valid JSON-RPC on stdout would catch regressions. Same for `additionalAllowedDirectories` — a test that loads a plugin from a non-default directory via config.

### 2. Concurrent stdout writes (LOW)

The notification SSE stream reader and the POST response handler both call `process.stdout.write()` concurrently from separate async contexts. Node's stream writes are not atomic for large payloads — two concurrent writes can interleave at the chunk level. A simple sequential write queue (or piping both through a shared transform) would make this safe.

### 3. SSE notification stream reconnection (LOW)

If the GET SSE notification stream drops (network hiccup, server restart during hot reload), it silently stops receiving notifications (`tools/list_changed`, logging). The bridge should detect the stream ending unexpectedly and reconnect with a small backoff, so the MCP client doesn't lose server-initiated messages for the rest of the session.

---

Everything else looks solid — architecture is correct, session lifecycle is clean, config preservation is thorough. Fix these three and it's good to go.

dathtd119:
> ## Review — Post-feedback commit ([fc5a80c](https://github.com/opentabs-dev/opentabs/commit/fc5a80c6d7b8d01a35f8921233f74743814a47f1))
> 
> The two MEDIUM issues (SSE detection, unbounded buffer) are fixed. Three items remain before merge:
> ### 1. E2E tests (MEDIUM)
> 
> The stdio bridge is a transport-layer feature that can regress silently. A basic Playwright/Vitest test that spawns `opentabs start --stdio`, writes `initialize` + `tools/list` on stdin, and asserts valid JSON-RPC on stdout would catch regressions. Same for `additionalAllowedDirectories` — a test that loads a plugin from a non-default directory via config.
> ### 2. Concurrent stdout writes (LOW)
> 
> The notification SSE stream reader and the POST response handler both call `process.stdout.write()` concurrently from separate async contexts. Node's stream writes are not atomic for large payloads — two concurrent writes can interleave at the chunk level. A simple sequential write queue (or piping both through a shared transform) would make this safe.
> ### 3. SSE notification stream reconnection (LOW)
> 
> If the GET SSE notification stream drops (network hiccup, server restart during hot reload), it silently stops receiving notifications (`tools/list_changed`, logging). The bridge should detect the stream ending unexpectedly and reconnect with a small backoff, so the MCP client doesn't lose server-initiated messages for the rest of the session.
> 
> Everything else looks solid — architecture is correct, session lifecycle is clean, config preservation is thorough. Fix these three and it's good to go.

Gotcha

PR #41, opened 2026-04-07
fix(ynab): align with YNAB sync API + add move_category_budget

tresni:
## Summary

Brings the YNAB plugin in line with how YNAB's internal `/api/v1/catalog` actually works. Multiple silent-failure bugs found and fixed during end-to-end testing against a live YNAB session.

### New tools
- **`move_category_budget`** — moves money between budget categories or to/from Ready to Assign for a specific month. Sends a `be_money_movements` record with the proper `manual_assign` / `manual_movement` source so the change actually persists.

### Read-side fixes
- `list_transactions` previously returned empty `payee_name`, `account_name`, `category_name` because YNAB's sync API only puts IDs on transactions — names live in separate `be_payees` / `be_accounts` / `be_subcategories` collections. Resolved via lookup maps.
- Field names were wrong: `accepted` (not `approved`), `flag` (not `flag_color`/`flag_name`), `imported_payee`/`original_imported_payee` (replacing the nonexistent `import_id`).
- Calc-map keying bugs in `get_month` and `list_categories` caused all category budgeted/activity to return zero — `buildSubcategoryCalcMap` was collapsing different months under the same key.
- `list_transactions` now supports `until_date` for date range queries.
- `cleared` and `flag_color` are now lowercased to match the documented schema.

### Write-side fixes
- Transaction writes now use `be_transaction_groups` (the format YNAB's UI sends). Without this wrapper, YNAB silently ignores the write.
- `update_category_budget` now sends a `be_money_movements` record alongside the budget change. Without it, YNAB reverts the budget on the next sync.
- Fixed direction handling: when lowering a budget, send positive amount with `from=category, to=null` (not negative amount with `to=category`).
- `create_transaction` resolves existing payees by name before creating duplicates.
- `update_transaction` and `delete_transaction` now refuse transfer and split transactions instead of silently corrupting them.
- `update_transaction` supports `flag_color: 'none'` to clear a flag.

### Quality / refactor
- Generic `syncWrite<T>` eliminates `BudgetEntities` casts across all write tools.
- Shared helpers in `schemas.ts`: `notTombstone`, `toMilliunits`, `toMonthKey`, `currentMonthKey`, `formatSubcategoryBudgetId`, `formatMonthlyBudgetId`, `MONEY_MOVEMENT_SOURCE`, `ClearedStatus`, `FlagColor`, `BudgetEntities`.
- 12 local `BudgetData` interfaces consolidated into one shared `BudgetEntities`.
- `mapPlan` JSON parse errors now logged via SDK `log.warn` instead of silently swallowed.
- `age_of_money` is nullable to distinguish "unknown" from "0".
- `assertAuthenticated()` helper for tools that need auth checks but not the plan ID.
- 426 sync error documented in memory for future debugging.

## Test plan

All 16 tools verified end-to-end against a live YNAB session with read-after-write checks for the write tools:

- [x] All 11 read tools return correct, populated data (names resolved, amounts correct)
- [x] `create_transaction` persists with payee resolution, flag, approved fields
- [x] `update_transaction` updates fields, clears flag with 'none', refuses transfers/splits
- [x] `delete_transaction` soft-deletes, refuses transfers
- [x] `update_category_budget` set positive delta — verified server state
- [x] `update_category_budget` set negative delta — verified server state
- [x] `move_category_budget` cat→cat — verified server state
- [x] `move_category_budget` RTA→cat — verified server state
- [x] `move_category_budget` cat→RTA — verified server state
- [x] Verified plugin still works after YNAB pushed 26.49.1-hotfix mid-session


opentabs-dev (commented):
## Deep Code Review

### Overall Assessment

Substantial and well-motivated PR. The core problem — YNAB's internal sync API uses different field names and wire formats than what the plugin originally assumed — is correctly identified and systematically fixed across all 16 tools. The refactoring from per-file `BudgetData` interfaces to a shared `BudgetEntities` type and the extraction of shared helpers into `schemas.ts` are clean improvements. The write-side fixes (transaction groups, money movements) are the kind of thing you can only get right by testing against the live API.

---

### Blocking (1)

**1. `syncWrite` reassigns a function parameter (`ynab-api.ts`)**

```ts
export const syncWrite = async <T = Record<string, unknown>>(
  planId: string,
  changedEntities: Record<string, unknown>,
  serverKnowledge?: number,   // <-- function parameter
): Promise<CatalogResponse<T>> => {
  if (serverKnowledge === undefined) {
    const readResult = await syncBudget(planId);
    serverKnowledge = readResult.current_server_knowledge ?? 0;  // <-- reassigned
  }
```

Parameter reassignment is flagged by Biome's `noParameterAssign` and is confusing — the caller's mental model is that the value they passed is what gets used. Use a local variable:

```ts
const knowledge = serverKnowledge ?? (await syncBudget(planId)).current_server_knowledge ?? 0;
```

---

### Important (4)

**2. Unsafe `as` cast drops `CatalogResponse` wrapper (`get-plan.ts`)**

```ts
const result = (await catalog('getInitialUserData', {
  device_info: { id: getDeviceId(), device_os: 'web' },
})) as InitialUserData;
```

`catalog` returns `Promise<CatalogResponse<T>>` which has `error`, `session_token`, `current_server_knowledge` fields. Casting the entire return value to `InitialUserData` silently discards these. If YNAB returns an error, `result.budget_version` would be `undefined` and the next check catches it — but the `error` field is now invisible. More precise fix:

```ts
const result = await catalog('getInitialUserData', { ... });
const budgetVersion = result.budget_version as RawPlan | undefined;
if (!budgetVersion) throw ToolError.notFound('No active plan found');
```

**3. `notTombstone` returns `boolean`, not a type predicate**

```ts
export const notTombstone = <T extends { is_tombstone?: boolean }>(x: T): boolean => !x.is_tombstone;
```

`.filter(notTombstone)` doesn't narrow the array type. Should use a type predicate:

```ts
export const notTombstone = <T extends { is_tombstone?: boolean }>(x: T): x is T & { is_tombstone?: false } =>
  !x.is_tombstone;
```

**4. `mapCategoryForMonth` drops `goal_type` from calc data**

The old code merged `goal_type` from the monthly calc record. The new `RawMonthlySubcategoryBudgetCalc` interface no longer has a `goal_type` field, so `mapCategoryForMonth` only uses `c.goal_type` (from the static subcategory definition). If YNAB stores month-specific goal changes on the calc record, historical months would show stale goals. If the removal is intentional because the YNAB API genuinely doesn't put `goal_type` on the calc, a brief comment would help.

**5. `currentMonthKey` uses browser local time, not YNAB budget timezone**

```ts
export const currentMonthKey = (): string => {
  const now = new Date();
  return `${now.getFullYear()}-${String(now.getMonth() + 1).padStart(2, '0')}`;
};
```

Around month boundaries, a timezone mismatch between the user's system and their YNAB budget could show last month's data. Minor edge case, not a showstopper.

---

### Minor (5)

**6. `buildLookups` maps empty-string IDs** — `id ?? ''` creates a phantom empty-string key. Transactions with no payee ID would match a random name. Filter out entries with no ID: `.filter(p => notTombstone(p) && p.id)`.

**7. `resolvePayee` creates untyped payload** — Returns `Record<string, unknown>` instead of typed `RawPayee`. Field name typos won't be caught at compile time.

**8. Hardcoded `cash_amount: 0, credit_amount: 0`** on transaction creates — may not be correct for credit card accounts. If tested and confirmed, a brief comment would help.

**9. `delete-transaction` spreads entire existing transaction** — Sends all raw fields back to the server. The old code was more surgical (only `id`, `entities_account_id`, `is_tombstone`). Could cause issues if the server rejects unknown/computed fields.

**10. `update-transaction` always sends `entities_scheduled_transaction_id: null`** — Severs the link to scheduled transactions. Should preserve existing value: `entities_scheduled_transaction_id: existing.entities_scheduled_transaction_id ?? null`.

---

### Positive

- The consolidation from 12 `BudgetData` interfaces into one `BudgetEntities` is the right call
- `buildLookups` resolving IDs to names from parallel collections is architecturally sound
- Transfer/split transaction guards prevent data corruption
- `resolvePayee` checking for existing payees before creating duplicates is a good UX improvement
- The `be_money_movements` discovery is critical — without it all budget writes silently revert
- `handleNetworkError` deduplication is clean
- The `payee_search` filter with pre-computed `matchingPayeeIds` set is a nice optimization

Great work tracking down these YNAB API quirks. The fix for the blocking item (#1) is a one-liner, and the important items are straightforward improvements. Looking forward to the next revision.

opentabs-dev (approved):
## Re-review — All feedback addressed

The latest commit (`1ac6180c`) addresses every finding from the initial review:

**Blocking (fixed):**
- `syncWrite` no longer reassigns `serverKnowledge` parameter — uses local `const knowledge` instead

**Important (all fixed):**
- `get-plan.ts`: Removed the unsafe `as InitialUserData` cast — now narrows only `budget_version` via the `CatalogResponse` index signature
- `notTombstone` is now a proper type predicate (`x is T & { is_tombstone?: false | undefined }`)
- `goal_type`: Comment confirms the calc record never carries `goal_type` on YNAB's wire format
- `currentMonthKey`: Comment explains the adapter runs inside the user's YNAB tab (same clock/timezone)

**Minor (all fixed):**
- `buildLookups`: New `hasId` type guard filters out entries with no ID — no more phantom empty-string key
- `resolvePayee`: Returns typed `RawPayee` instead of `Record<string, unknown>`; `RawPayee` interface expanded with auto-fill fields
- `cash_amount`/`credit_amount`: Documented what's verified
- `delete-transaction`: Comment explaining YNAB rejects minimal payloads (400)
- `update-transaction`: Preserves existing `entities_scheduled_transaction_id` instead of always sending `null`

Clean, thorough response to review. LGTM — approving.

PR #62, opened 2026-04-19
feat(mcp-server): emit browser_screenshot_tab as MCP image content part

raffishquartan:
## Summary

Fixes an inconsistency in `browser_screenshot_tab`'s response shape. Today the tool's base64 PNG is JSON-stringified into a `text` content part. Small results arrive inline as `[{type:'text', text:'{\"image\":...}'}]`; large results are spilled by Claude Code (and likely other clients) to a tool-result file that contains only the parsed inner JSON `{image:...}` - no content-array wrapper. Callers have to branch on the input type.

The fix emits a proper MCP `image` content part (`{type:'image', data:'<base64>', mimeType:'image/png'}`) instead. Clients decode image content parts natively regardless of size, so the text-spill path is bypassed entirely and consumer parse paths become uniform.

### Implementation

- `BrowserToolDefinition` gains an optional `formatResult?: (result) => ToolContentPart[]` hook. The dispatcher uses it when defined; otherwise it falls back to the existing JSON-stringify-as-text wrapping. This keeps the door open for any future browser tool that needs to return image/audio/binary payloads.
- `browser_screenshot_tab` ships its own `formatResult` that transforms the extension's `{image: <base64>}` into the single-part image array.
- `ToolCallResult.content` is widened from `Array<{type:'text', text:string}>` to `ToolContentPart[]` (text | image). Downstream test assertions move onto a small `textPart` narrow-helper.
- Tool description, `BROWSER_TOOLS_CATALOG`, and the docs entry updated to reflect the new return shape.
- `platform/cli/src/commands/tool.ts` is updated to render image content parts (and report unsupported/malformed parts) instead of silently dropping them. Image parts are written to `os.tmpdir()` and the resulting path is reported on stdout alongside the mimeType and approximate byte count.

No behaviour change for any tool other than `browser_screenshot_tab`.

### Test plan

- [x] `tsc --build` clean
- [x] `npm run lint` clean
- [x] `npm run format:check` clean
- [x] `npm run knip` clean
- [x] `npm run generate:browser-tools-catalog` reports up-to-date
- [x] 1593 mcp-server + shared unit tests pass
- [x] 344 CLI unit tests pass (including 5 new `renderToolCallContent` tests)
- [x] New unit test: `handleBrowserToolCall` routes through `formatResult` when a tool defines it
- [x] New unit test: `screenshotTab.formatResult` emits a single-part `image/png` content block
- [x] Live end-to-end: rebuilt the server, restarted the daemon, called `browser_screenshot_tab` via a raw HTTP MCP client. Response was `{content: [{type:'image', mimeType:'image/png', data:<valid-PNG-base64>}]}`; PNG magic bytes `89504e470d0a1a0a` present; no `isError`.

[Cld]


README-HOWTO-UPDATEFORK-TO-MYSTREAM.md
```diff
@@ -0,0 +1,178 @@
+# How to update this fork from upstream opentabs
```
opentabs-dev, line 1: Drop this file. It's a personal fork-management guide — it references your fork name (`raffishquartan/opentabs-rq`), your local paths (`~/repos/opentabs-rq`), and your private Claude Code hook (`~/repos/claude-code-config-sync/hooks/session-start/start-opentabs.sh`). None of this belongs on `main`.

.gitignore
```diff
@@ -25,3 +25,4 @@ marketing/logs/
 marketing/state.json
 .playwright-browsers/
 .extension-hash
+docs/superpowers/
```
opentabs-dev, line 28: Drop `docs/superpowers/` — that's a local plugin-cache directory on your machine, not an opentabs artifact. Should live in your global gitignore, not the repo.

platform/mcp-server/src/browser-tools/add-tabs-to-group.ts
```diff
@@ -0,0 +1,29 @@
...
+  group: 'Tabs',
+  input: z.object({
+    groupId: z.number().int().nonnegative().describe('Target group ID. Use browser_list_tab_groups to find IDs.'),
+    tabIds: z.array(z.number().int().positive()).min(1).describe('Tab IDs to add to the group.'),
+  }),
+  handler: async (args, state) =>
+    dispatchToExtension(state, 'browser.addTabsToGroup', {
+      groupId: args.groupId,
+      tabIds: args.tabIds,
+      tabId: args.tabIds[0],
```
opentabs-dev, line 25: 🔴 `tabId: args.tabIds[0]` is a routing hack — don't do this. `tabIds[]` can span multiple profiles; picking index 0 silently decides which profile gets the dispatch. Other IDs in the array from different profiles will be rejected by Chrome with no useful error.

Fix: add an optional `connectionId` input (consistent with `update-tab-group.ts` / `list-tabs-in-group.ts`) and route via that. `resolveConnection` at `extension-protocol.ts:183-187` already prioritizes explicit `connectionId` over `tabId`-based resolution. Remove the synthetic `tabId` from the params.

platform/mcp-server/src/browser-tools/create-tab-group.ts
```diff
@@ -0,0 +1,38 @@
...
+    title: z.string().optional().describe('Display name for the tab group'),
+    color: z
+      .enum(TAB_GROUP_COLORS)
+      .optional()
+      .describe('Tab group color. Defaults to Chrome-assigned color if omitted.'),
+  }),
+  handler: async (args, state) =>
+    dispatchToExtension(state, 'browser.createTabGroup', {
+      tabIds: args.tabIds,
+      tabId: args.tabIds[0],
```
opentabs-dev, line 32: 🔴 Same issue as `add-tabs-to-group.ts:25` — drop the synthetic `tabId`, add optional `connectionId`, route explicitly. `tabIds[]` that span profiles will silently route to whichever profile owns the first one.

Also note the API asymmetry: `update-tab-group` and `list-tabs-in-group` take `connectionId` but `create-tab-group`, `add-tabs-to-group`, and `remove-tabs-from-group` don't. Make all six consistent.

platform/mcp-server/src/browser-tools/remove-tabs-from-group.ts
```diff
@@ -0,0 +1,30 @@
...
+  input: z.object({
+    tabIds: z
+      .array(z.number().int().positive())
+      .min(1)
+      .describe('Tab IDs to remove from their current group (ungroup).'),
+  }),
+  handler: async (args, state) =>
+    dispatchToExtension(state, 'browser.removeTabsFromGroup', {
+      tabIds: args.tabIds,
+      tabId: args.tabIds[0],
```
opentabs-dev, line 26: 🔴 Same `tabId: args.tabIds[0]` hack — fix by exposing `connectionId` and dropping the synthetic param. See the comment on `add-tabs-to-group.ts:25`.

platform/mcp-server/src/browser-tools/list-tabs-in-group.ts
```diff
@@ -0,0 +1,40 @@
...
+      .optional()
+      .describe(
+        'Target a specific browser profile. Get from browser_list_tab_groups. ' +
+          'Required in multi-profile setups so the query reaches the profile that owns the group.',
+      ),
+  }),
+  handler: async (args, state) =>
+    dispatchToExtension(state, 'browser.listTabsInGroup', {
+      groupId: args.groupId,
+      ...(args.connectionId ? { connectionId: args.connectionId } : {}),
```
opentabs-dev, line 36: Use `args.connectionId !== undefined` instead of the truthy check — Zod's `z.string().optional()` allows empty string, and `args.connectionId ? ...` silently drops `""`. `update-tab-group.ts:47` already uses the correct form; make this match. CodeRabbit flagged this.

platform/mcp-server/src/browser-tools/screenshot-tab.ts
```diff
@@ -23,6 +24,15 @@ const screenshotTab = defineBrowserTool({
       .describe('Tab ID to screenshot — the tab will be focused automatically before capture'),
   }),
   handler: async (args, state) => dispatchToExtension(state, 'browser.screenshotTab', { tabId: args.tabId }),
+  formatResult: result => {
+    const data = (result as { image?: unknown } | null)?.image;
+    if (typeof data !== 'string') {
+      throw new Error(
```
opentabs-dev, line 30: Don't serialize the full payload into the error message — `JSON.stringify(result)` could leak arbitrary bytes the extension returned (in the current shape it's a base64 PNG, but future extension changes or bugs could put anything there) into an MCP error the client logs.

Replace with safe metadata:

```ts
throw new Error(
  `browser_screenshot_tab: extension returned unexpected payload (expected { image: string }, got keys: ${Object.keys(result ?? {}).join(', ') || typeof result})`,
);
```

CodeRabbit flagged this.

platform/mcp-server/src/browser-tools/screenshot-tab.test.ts
```diff
@@ -0,0 +1,10 @@
+import { describe, expect, test } from 'vitest';
+import { screenshotTab } from './screenshot-tab.js';
+
+describe('screenshotTab.formatResult', () => {
+  test('emits a single MCP image content part with mimeType image/png', () => {
+    expect(screenshotTab.formatResult).toBeDefined();
+    const formatted = screenshotTab.formatResult?.({ image: 'iVBORw0KGgoAAAANSUhEUg==' });
+    expect(formatted).toEqual([{ type: 'image', data: 'iVBORw0KGgoAAAANSUhEUg==', mimeType: 'image/png' }]);
+  });
```
opentabs-dev, line 9: Add a failure-case test alongside the success one — pass a malformed payload (e.g. `{ image: 123 }` or `null`) and assert `formatResult` throws with the expected message. That locks in the runtime guard at `screenshot-tab.ts:27-30`. CodeRabbit suggested the exact diff.

platform/cli/src/commands/tool.ts
```diff
@@ -171,22 +224,35 @@ const handleToolCall = async (
 
   const result = (await res.json()) as ToolCallResult;
 
-  // Extract text content from the result
-  const texts = result.content.filter(c => c.type === 'text').map(c => c.text);
-  const output = texts.join('\n');
+  // Save image content parts to tmpdir so their paths can be reported to the user
+  const saveImage = (data: string, mimeType: string, index: number): string => {
+    const ext = extForMime(mimeType);
+    const filename = `opentabs-${name}-${Date.now()}-${index}.${ext}`;
```
opentabs-dev, line 230: `writeFileSync` here is unwrapped — `ENOSPC`, `EACCES`, or a tmpdir that doesn't exist crashes the CLI mid-render with a raw stack trace. Wrap it and route the failure through `renderToolCallContent` as a `[image: save failed — <reason>]` line so the user sees which image couldn't be saved and keeps the rest of the output.

CodeRabbit flagged this.

opentabs-dev (changes requested):
Thanks for the screenshot work — the `formatResult` hook + image-content-part design is the right MCP-native fix and I want to merge it. But I can't take this PR as-is. Asking for changes at the scope level before we dig deeper.

## 🔴 Blocker 1 — Split this PR

The title and description say "No behaviour change for any tool other than `browser_screenshot_tab`", but the diff adds:

- 6 new MCP browser tools for Chrome tab-groups (`browser_list_tab_groups`, `browser_create_tab_group`, `browser_add_tabs_to_group`, `browser_remove_tabs_from_group`, `browser_update_tab_group`, `browser_list_tabs_in_group`)
- a new `tabGroups` Chrome manifest permission
- a new shared `tab-group-colors` module + subpath export
- ~340 LOC of handler + dispatch tests

That's a real feature. It deserves its own PR with its own threat model and its own review. Please split into:

1. **Tab-groups feature** (commits `f8faac4f`..`05b96375`) — separate PR, rewritten description that describes the actual change, with the issues below resolved.
2. **`formatResult` hook + screenshot + CLI image rendering** (`f7535224`..`22612118`) — keep this PR focused on that. I'll merge it fast once the fork-personal files are dropped and CodeRabbit's CLI/test nits are addressed.

## 🔴 Blocker 2 — Fork-personal files

- `README-HOWTO-UPDATEFORK-TO-MYSTREAM.md` — this is a personal fork-management guide with your paths (`~/repos/opentabs-rq`, `raffishquartan/opentabs-rq`, `~/repos/claude-code-config-sync/hooks/session-start/start-opentabs.sh`). It doesn't belong upstream.
- `.gitignore` `docs/superpowers/` entry — that's a local plugin-cache directory on your machine, not an opentabs concern.

Please remove both before this lands.

## 🔴 Blocker 3 — `tabId: args.tabIds[0]` routing hack (tab-groups PR)

In `add-tabs-to-group.ts`, `create-tab-group.ts`, and `remove-tabs-from-group.ts` the handler injects `tabId: args.tabIds[0]` into the dispatch params to piggy-back on `resolveConnection`'s tab-based routing. This is wrong:

- `tabIds[]` can span multiple profiles — picking index 0 silently decides routing. Any tab IDs from other profiles produce a confusing Chrome error with no hint about why.
- The documented convention (CLAUDE.md: "Plugin tool handlers never see `tabId` in their input — it is a platform concern") is that `tabId` semantics belong to the platform, not synthesized at the call site.
- `resolveConnection` already supports explicit `connectionId` as the first-priority branch (`extension-protocol.ts:183-187`). Use that.

Fix: expose `connectionId` on all three tools consistently (see `update-tab-group.ts`/`list-tabs-in-group.ts`, which already do) and drop the synthetic `tabId`. Or compute `connectionId` from `getConnectionForTab(state, tabIds[0])` server-side and pass it explicitly — whatever you do, don't tunnel through `tabId`.

## 🟡 API asymmetry across the six new tools

`update-tab-group` and `list-tabs-in-group` accept `connectionId`. `create-tab-group`, `add-tabs-to-group`, `remove-tabs-from-group` don't. Multi-profile agents can target a profile for reads/updates but not for creates/adds/removes. Pick one behavior and apply it to all six.

## 🟡 `dispatchToAllConnections` partial-failure silence

`browser_list_tab_groups` uses `dispatchToAllConnections`, which drops failed profiles silently. Agent gets an incomplete list with no signal that profile X failed. Follow-up OK, but at least document the choice in the tool description or surface failed connectionIds in the result envelope.

## 🟡 Mock reset drift (outside diff range, so flagging here)

`resetRoutingMocks()` in `platform/browser-extension/src/message-router.test.ts` around line 733 doesn't reset the six new tab-group handler mocks (`mockHandleBrowserListTabGroups`, `mockHandleBrowserCreateTabGroup`, `mockHandleBrowserAddTabsToGroup`, `mockHandleBrowserRemoveTabsFromGroup`, `mockHandleBrowserUpdateTabGroup`, `mockHandleBrowserListTabsInGroup`). Order-dependent test leak — CodeRabbit flagged it too, has the exact diff.

## 🟡 CodeRabbit findings — please address inline

I'm leaving inline comments on the `tabId` hacks, the `writeFileSync` crash in the CLI, the payload-leaking error in `screenshot-tab.ts`, the empty-string `connectionId` guard, and the missing failure-case test. Several overlap with CodeRabbit's comments — please resolve both.

## 🟢 What's good

- `formatResult` hook is the right minimal extension to `BrowserToolDefinition`. Default text-wrap path unchanged → zero risk to other tools.
- `tab-group-colors.ts` in `@opentabs-dev/shared` with side-effect-free subpath export is correctly placed.
- The `textPart` narrow-helper pattern in the two test files is how I want union-typed content asserted going forward.
- Live end-to-end verification with a raw MCP client in the test plan is exemplary — keep doing that.

## What I want before re-review

1. Split into two PRs (or drop to just the screenshot one + a separate tab-groups PR).
2. Delete the fork-personal files.
3. Screenshot PR: fix `saveImage` crash + tighten `screenshot-tab.ts` error payload + add malformed-payload test. That PR I'll merge quickly.
4. Tab-groups PR: fix `tabIds[0]` routing hack across the three handlers, add `connectionId` to the missing three, fix empty-string guard, add mock resets, rewrite description to reflect the feature + the new manifest permission.
5. Rebase onto main — PR is currently `CONFLICTING`.

raffishquartan:
Superseded by #66 — rebased cleanly onto upstream/main, tab-groups and fork-personal files removed, reviewer feedback applied (payload-leak in error, malformed-payload test, CLI writeFileSync safety). Please continue review there.

Issue #64, opened 2026-04-19
outlook adapter blocked by Trusted Types policy — TrustedScript assignment fails on Outlook Web App


conceptblenders:
## Summary
After the `cloud.microsoft` domain fix in 0.0.103, the adapter now runs but
immediately throws a Trusted Types violation. The plugin cannot initialize.

## Error
"This document requires 'TrustedScript' assignment. The action has been blocked."
Source: adapters/outlook-63c4a43f.js:1352

## Environment
- @opentabs-dev/opentabs-plugin-outlook: 0.0.89
- @opentabs-dev/cli: 0.0.103
- OS: Windows 11, Chrome
- URL: https://outlook.cloud.microsoft/mail/

## Root cause
Microsoft Outlook Web App enforces a strict Trusted Types CSP policy.
The adapter uses an untrusted script assignment (innerHTML, eval(), or
new Function()) that is blocked by this policy.

## Fix required
The adapter needs to use the Trusted Types API:

  const policy = trustedTypes.createPolicy('opentabs', {
    createScript: (s) => s,
    createHTML: (s) => s,
  });

Or avoid eval/innerHTML patterns entirely and use DOM APIs directly
(document.createElement, textContent, addEventListener).

## Steps to reproduce
1. Install @opentabs-dev/opentabs-plugin-outlook
2. Open https://outlook.cloud.microsoft/mail/ in Chrome with OpenTabs extension
3. Observe TrustedScript error in console — adapter does not become ready

opentabs-dev:
Thanks for the detailed report! Unfortunately I can only reproduce against `outlook.live.com` (consumer Outlook) — I don't have a Microsoft 365 work/school tenant, so `outlook.cloud.microsoft` redirects me away and I can't exercise the exact CSP policy that's triggering the violation.

A few notes from what I could dig into locally:

- Line 1352 of the built adapter points at `new Function("")` inside zod's internal `allowsEval` feature-detect probe. That probe is already wrapped in try/catch, so the throw itself should be non-fatal — zod falls back to its non-eval path when `allowsEval` returns false.
- Could you confirm whether the adapter eventually becomes `ready` despite the console error? You can check via `extension_check_adapter` (look for `isReady: true` on the matching tab), or just see whether plugin tools work on that tab after the error appears.
[...]

PR #65, opened 2026-04-20
fix(outlook): add Trusted Types policy and extend auth timeout

conceptblenders:
Outlook on cloud.microsoft enforces a Trusted Types CSP policy that causes zod's allowsEval probe (new Function("")) to log a console violation. Creating a 'default' Trusted Types policy before any zod code runs lets the probe complete silently via its existing try/catch.

Also extends waitForAuth timeout from 5000ms to 30000ms. When Outlook redirects through an OAuth flow (#code= URL fragment), MSAL tokens are not yet in localStorage when the adapter first initializes. The previous 5s window was too short for the OAuth handshake to complete on slower connections or enterprise SSO flows.

Fixes two issues reported in [issue #<nummer>] affecting Outlook on `cloud.microsoft`:

### 1. Trusted Types policy (`index.ts`)
Outlook enforces a Trusted Types CSP. Zod's `allowsEval` probe calls
`new Function("")` which triggers a console violation. Creating a `default`
Trusted Types policy before any zod code runs lets the probe complete
silently via its existing try/catch.

### 2. Extended auth timeout (`outlook-api.ts`)
`waitForAuth` timeout increased from 5000ms → 30000ms.

When Outlook loads via an OAuth redirect (`#code=` URL fragment), MSAL tokens
are not yet in localStorage when the adapter first initializes. The 5s window
was too short for the OAuth handshake + token storage to complete, causing
`isReady()` to return false permanently.

## Test environment
- Microsoft 365 work/school account (enterprise tenant)
- `https://outlook.cloud.microsoft/mail/`
- Chrome + OpenTabs extension v0.0.103

## Summary by CodeRabbit

* **New Features**
  * Adds a runtime Trusted Types policy to tighten script handling and avoid startup interruptions if policy setup fails.

* **Bug Fixes**
  * Defer token cache checks during the OAuth redirect phase to prevent premature authentication failures and improve sign-in reliability.

### DIFF
--- a/plugins/outlook/src/index.ts
+++ b/plugins/outlook/src/index.ts
@@ -4,8 +4,18 @@
 // succeed silently instead of logging a console violation.
 if (typeof window !== 'undefined') {
   try {
-    const tt = (window as unknown as { trustedTypes?: { createPolicy?: (name: string, rules: Record<string, (s: string) => string>) => void } }).trustedTypes;
-    tt?.createPolicy?.('default', { createScript: (s: string) => s });
+    const tt = (window as Window & {
+      trustedTypes?: TrustedTypePolicyFactory & { defaultPolicy?: TrustedTypePolicy | null };
+    }).trustedTypes;
+    if (tt && !tt.defaultPolicy) {
+      tt.createPolicy('default', {
+        createScript: (s: string) => {
+          // Only allow the empty-string probe used by zod's allowsEval feature-detect.
+          if (s !== '') throw new TypeError('Blocked Trusted Types script conversion');
+          return s;
+        },
+      });
+    }
   } catch {
     // 'default' policy already exists, or Trusted Types not supported — safe to ignore.
   }
--- a/plugins/outlook/src/index.ts
+++ b/plugins/outlook/src/index.ts
@@ -1,7 +1,8 @@
-// Outlook enforces Trusted Types (CSP). Zod's allowsEval probe calls
-// new Function("") which requires a 'default' Trusted Types policy to exist.
-// Creating it here (before any zod code runs) lets the try/catch in zod
-// succeed silently instead of logging a console violation.
+// Outlook on cloud.microsoft enforces Trusted Types (CSP). Zod's allowsEval
+// feature-detect calls new Function("") the first time a zod parser runs
+// (lazy, memoized via cached()). This policy just needs to exist before that
+// first runtime call — import statements are hoisted regardless, but the probe
+// only fires when a tool handler first invokes a zod schema at runtime.
 if (typeof window !== 'undefined') {
   try {
     const tt = (window as Window & {
--- a/plugins/outlook/src/index.ts
+++ b/plugins/outlook/src/index.ts
@@ -1,3 +1,16 @@
+// Outlook enforces Trusted Types (CSP). Zod's allowsEval probe calls
+// new Function("") which requires a 'default' Trusted Types policy to exist.
+// Creating it here (before any zod code runs) lets the try/catch in zod
+// succeed silently instead of logging a console violation.
+if (typeof window !== 'undefined') {
+  try {
+    const tt = (window as unknown as { trustedTypes?: { createPolicy?: (name: string, rules: Record<string, (s: string) => string>) => void } }).trustedTypes;
+    tt?.createPolicy?.('default', { createScript: (s: string) => s });
+  } catch {
+    // 'default' policy already exists, or Trusted Types not supported — safe to ignore.
+  }
+}
+
 import { OpenTabsPlugin } from '@opentabs-dev/plugin-sdk';
 import type { ToolDefinition } from '@opentabs-dev/plugin-sdk';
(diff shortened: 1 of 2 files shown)


opentabs-dev (changes requested):
Thanks for the PR — appreciate the investigation. A few blocking concerns before I can merge:

**1. The waitForAuth timeout bump to 30s is a no-op.**

The platform caps adapter `isReady()` probes at 5 seconds (`platform/browser-extension/src/constants.ts:34`, `IS_READY_TIMEOUT_MS = 5_000`). If `isReady()` doesn't resolve within that window, `computePluginTabState` marks the tab as `unavailable` regardless of what the adapter is still waiting for. So a 30s timeout inside the plugin never actually runs — the probe is already dead at 5s.

The readiness polling re-probes every 30s (`READINESS_POLL_INTERVAL_MS`), so a slow OAuth handshake is already handled: the first probe fails at 5s, and subsequent probes catch the token once it lands. Bumping to 30s doesn't help and gives a misleading signal about what the platform guarantees.

If the real issue is 'first probe too early after OAuth redirect', then either (a) fix it in `isAuthenticated()` itself (detect the `#code=` fragment and return early with false, letting the poll re-probe), or (b) open a separate discussion about raising `IS_READY_TIMEOUT_MS` — that's a platform-level change, not a plugin one. Either way, please drop the `outlook-api.ts` change from this PR.

**2. The Trusted Types policy is too permissive.**

CodeRabbit flagged this, and I agree: installing a `default` Trusted Types policy whose `createScript` returns the input unconditionally would — if it were actually reached — disable the page's script-injection defenses for anything else that tries to assign a script. Your current commit already restricts it to empty strings, which is good. Keep that restriction.

**3. The 'before any zod code runs' comment is incorrect.**

ES module `import` statements are hoisted: the imports on lines 25–41 execute before the top-level statement block you added above them. So if zod's `allowsEval` probe runs at module-eval time through one of those imports, your policy is installed *after* zod has already run. It happens to work here because zod's `allowsEval` is a memoized lazy call (`cached(...)`) that only fires the first time it's invoked at runtime, not at import time — but the comment suggests ordering that the language doesn't provide.

Please either (a) move the Trusted Types setup into a separate module imported first and explain *why* it works (lazy probe, not import-time), or (b) rewrite the comment to accurately describe the mechanism: 'zod's allowsEval probe is lazy — this policy just needs to exist before the first runtime call into a zod parser.'

**4. No test environment confirmation.**

I can't reproduce this without a Microsoft 365 tenant. Before merging, can you confirm on `outlook.cloud.microsoft`:
- Does `extension_check_adapter` report `isReady: true` after the fix?
- Does at least one plugin tool (e.g., `outlook_get_current_user`) return successfully?

A screenshot or MCP output in this PR would be enough.

Summary: split the auth-timeout change out (it doesn't do what the description claims), tighten the comment to reflect lazy probe semantics, and share a minimal repro showing tools work. Happy to merge once those are addressed.

opentabs-dev:
Thanks for finding and surfacing both of these — the zod/`new Function("")` probe interaction with Trusted Types is a subtle one, and the OAuth-redirect timing issue is real.

We'd like this to land on the pre-script SDK we shipped yesterday: https://opentabs.dev/docs/sdk/utilities#pre-script-utilities

Both fixes in this PR have correctness gaps that pre-scripts close:

**Trusted Types policy in `index.ts` is too late.** Bundlers hoist/reorder imports, so any zod import above your policy-creation line (directly or transitively) will run first — and if `allowsEval` fires during that import, it's blocked before the policy exists. A pre-script runs at `document_start` in MAIN world, strictly before any page script and before the adapter IIFE's own imports execute, so the policy is guaranteed to exist when zod probes:

```typescript
// plugins/outlook/src/pre-script.ts

export default definePreScript(() => {
  if (typeof trustedTypes !== 'undefined' && !trustedTypes.defaultPolicy) {
    trustedTypes.createPolicy('default', {
      createScript: (s) => s,
      createHTML: (s) => s,
      createScriptURL: (s) => s,
    });
  }
});
```

**The `waitForAuth` 5s → 30s extension is a workaround for a different problem** — "MSAL hasn't written to localStorage yet after OAuth redirect." If you also intercept tokens from `window.fetch` in the same pre-script, you don't need to wait for MSAL at all. The token is captured the instant Outlook makes its first authenticated request:

```typescript
// Same file, add:
const real = window.fetch;
  const url = typeof input === 'string' ? input : input instanceof Request ? input.url : input.toString();
  if (auth && /(outlook\.office\.com|graph\.microsoft\.com)/.test(url)) {
    set('authToken', auth);
  }
  return real(input, init);
};
```

Adapter reads with `getPreScriptValue<string>('authToken')`. `plugins/prescript-test/` is a working reference.

One honest tradeoff: pre-scripts only fire on future navigations, so on first plugin install a user with Outlook already open needs to reload once. After that, plugin updates auto-reload matching tabs on `preScriptHash` changes.

This PR plus #64 (the adapter TrustedScript error) plus #69 (cloud.microsoft token interception) all converge on the same pre-script file. Worth doing once, right. Happy to review a rework.

PR #68, opened 2026-04-22
fix(slack): use V1 files.upload API to avoid CORS failure

### LINKED ISSUE #67: Slack file upload fails with CORS error (files.uploadV2 cross-origin fetch)
## Problem

`slack_upload_file` always fails with a network error:

```
upload-file: network error during file upload: Failed to fetch
{"code":"network_error","category":"internal","retryable":true}
```

This happens for every upload, regardless of file size, content type, or channel. All other Slack tools (send_message, list_files, read_messages, etc.) work fine.

## Root cause

The upload tool uses Slack's V2 upload API, which is a three-step process:

1. `files.getUploadURLExternal` — returns an upload URL (same-origin call to `app.slack.com/api/...`, succeeds)
2. `fetch(upload_url, ...)` — POSTs file bytes to the returned URL (cross-origin call to `files.slack.com`, **fails**)
3. `files.completeUploadExternal` — finalizes the upload (never reached)

The adapter runs in the page's `MAIN` world (injected via `chrome.scripting.executeScript` with `world: 'MAIN'`). Step 2 fails because:

- The upload URL is on `files.slack.com`, a different origin from `app.slack.com`
- The page's same-origin policy blocks the cross-origin fetch
- `credentials: 'include'` makes CORS preflight requirements stricter (needs exact origin match)
- `redirect: 'error'` causes an immediate `TypeError: Failed to fetch` if the endpoint redirects

The extension's `manifest.json` has `<all_urls>` in host_permissions, but those only apply to the background service worker, not to scripts running in `MAIN` world.

## Why other tools work

`slackApi()` sends all requests to `${auth.workspaceUrl}/api/${method}` which resolves to `https://app.slack.com/api/...`. This is same-origin from the MAIN world context. The file upload is the only tool that makes a cross-origin request.

## Proposed fix

Use Slack's V1 `files.upload` API instead of the V2 multi-step flow. The V1 API accepts file content inline as part of a single multipart form POST to `app.slack.com/api/files.upload`, keeping everything same-origin.

While `files.upload` V1 is deprecated, it remains functional and is the only viable approach without SDK-level changes to support background-proxied fetches.

A longer-term fix would be to add a `backgroundFetch` capability to the plugin-sdk that routes cross-origin fetches through the extension's background service worker (which has `<all_urls>` privileges and isn't subject to page CORS).

## Relevant files

- `plugins/slack/src/tools/upload-file.ts` (lines 76-123) — the upload implementation
- `plugins/slack/src/slack-api.ts` — the `slackApi()` function (same-origin, works)
- `platform/browser-extension/src/tool-dispatch.ts` — tools execute in MAIN world
- `platform/browser-extension/src/iife-injection.ts` — adapter injection into MAIN world

## Environment

- OpenTabs `@opentabs-dev/opentabs-plugin-slack` (latest npm)
- macOS, Chrome, Slack open at `app.slack.com`
- Reproducible on every call, not intermittent

rbutera:
## Summary

`slack_upload_file` fails on every call with `Failed to fetch`. The V2 upload API needs a cross-origin fetch to `files.slack.com`, which the browser blocks because the adapter runs in MAIN world on `app.slack.com`.

This switches to Slack's V1 `files.upload` API, posting the file content as a multipart form to the same-origin `/api/files.upload` endpoint. Every other Slack tool already uses same-origin requests through `slackApi()`. This brings file uploads in line.

### What changed

- **`plugins/slack/src/tools/upload-file.ts`**: Replaced the three-step V2 upload flow with a single multipart form POST to `files.upload` (V1). Added rate-limit handling and response parsing for the V1 response shape.
- **`plugins/slack/src/slack-api.ts`**: Exported `getAuth()` so the upload tool can build its own authenticated multipart request.

### Trade-offs

The V1 `files.upload` API is deprecated by Slack but still works. The V2 API can't work from MAIN world without SDK-level support for routing fetches through the background service worker (which has `<all_urls>` privileges). A `backgroundFetch` SDK primitive would be the proper long-term fix, but this unblocks file uploads now.

Fixes #67

## Test plan

- [ ] Upload a text file to a DM channel
- [ ] Upload a text file to a public channel
- [ ] Upload a base64-encoded image to a channel
- [ ] Confirm blocked extensions (.exe, .sh, etc.) still reject
- [ ] Confirm other Slack tools still work (send_message, read_messages, list_files, etc.)

## Summary by CodeRabbit

* **New Features**
  * Improved Slack authentication support for integrations, enabling smoother sign-in and workspace linkage.

* **Bug Fixes**
  * More reliable Slack file uploads with better handling of missing auth, timeouts, rate limits, invalid responses, and detailed error mapping.

### DIFF
--- a/plugins/slack/src/slack-api.ts
+++ b/plugins/slack/src/slack-api.ts
@@ -508,4 +508,4 @@ const slackApi = async <T extends Record<string, unknown>>(
   return data as T & { ok: true };
 };
 
-export { isSlackAuthenticated, waitForSlackAuth, slackApi };
+export { getAuth, isSlackAuthenticated, waitForSlackAuth, slackApi };
--- a/plugins/slack/src/tools/upload-file.ts
+++ b/plugins/slack/src/tools/upload-file.ts
@@ -1,5 +1,5 @@
 import { getAuth } from '../slack-api.js';
-import { ToolError, defineTool } from '@opentabs-dev/plugin-sdk';
+import { ToolError, defineTool, parseRetryAfterMs } from '@opentabs-dev/plugin-sdk';
 import { z } from 'zod';
 
 export const uploadFile = defineTool({
@@ -73,10 +73,7 @@ export const uploadFile = defineTool({
       }
     }
 
-    // Use the V1 files.upload API via a same-origin multipart form POST.
-    // The V2 flow (getUploadURLExternal + fetch to files.slack.com +
-    // completeUploadExternal) fails because the adapter runs in MAIN world
-    // and the upload URL is on a different origin, blocked by CORS.
+    // V1 files.upload: same-origin multipart form POST via getAuth() credentials.
     const auth = getAuth();
     if (!auth) {
       throw ToolError.auth('Not authenticated — no Slack session token found');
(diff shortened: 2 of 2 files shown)


opentabs-dev:
Thanks for the thorough fix and clear root-cause writeup @rbutera — this was a clean diagnosis.

Merged to `main` as `6fdab301`. Two small follow-ups were pushed to your branch before merge, both addressing CodeRabbit nits:

- Normalized the 429 retry hint to derive from parsed `retryMs` rather than the raw `Retry-After` header, since the header may be an RFC 7231 HTTP-date rather than seconds.
- Mapped HTTP 400 to `ToolError.validation` (it indicates malformed input, not a server failure).

Admin-merged because of a repo-side config issue unrelated to your PR: the `check-docs` required status check is path-filtered to `docs/**` and never reports on non-docs PRs, so the merge was structurally blocked. Fixing that workflow in a follow-up.

For the longer-term direction you mentioned in the PR description — a `backgroundFetch` SDK primitive that routes cross-origin requests through the background service worker — that's the right architecture. If you're interested in tackling it, worth opening a separate issue first to discuss the API shape.

PR #72, opened 2026-04-24
fix(ynab): preserve imported_date on update to stop duplicate imports

tresni:
## Summary

- `update_transaction` reconstructed the transaction row and hardcoded `imported_date`, `matched_transaction_id`, `provider_cleansed_payee`, and other import-metadata fields to `null` on every call. The YNAB server silently accepted the null for `imported_date`, breaking bank-feed dedup and causing duplicate imports on heavily-edited accounts (anecdotally clustered around edit activity on Chase).
- Fix: spread `...existing` so every wire field round-trips untouched, matching the UI's full-replace pattern. Only explicitly-requested fields are overridden.
- Widen `RawTransaction` to carry the full wire schema so `existing` actually has those fields to preserve.
- Surface `imported_date`, `ynab_id`, `matched_transaction_id`, and `source` on `list_transactions` / `get_transaction` output so future import-dedup investigations don't require raw `/api/v1/catalog` calls.

## Investigation trail

Captured the UI's actual `syncBudgetData` payload for updates, then did a controlled experiment on a live auto-imported transaction: sent an update through our tool and re-read the raw server state. `imported_date` went from `"2026-04-22"` to `null` — confirmed persistent mutilation. Other fields we blanked (`cash_amount`, `credit_amount`, `credit_amount_adjusted`, `subcategory_credit_amount_preceding`, `provider_cleansed_payee`) were either recomputed or restored by the server, so the narrow fix is `imported_date`-and-siblings via spread, not every hardcoded null.

Re-ran the same experiment post-fix — `imported_date` now round-trips cleanly.

## Test plan

- [x] `npm run build` (plugin)
- [x] `npm run type-check`
- [x] `npm run lint`
- [x] `npm run format:check`
- [x] Live verification: updated a real auto-imported Chase transaction via the fixed tool and confirmed `imported_date`, `ynab_id`, `source`, `provider_cleansed_payee`, and `matched_transaction_id` all preserved on re-read via raw `syncBudgetData`

## Notes

- Does **not** retroactively merge duplicates that already exist; those need to be handled manually in the UI.
- `provider_cleansed_payee` turned out to be server-restored (not server-persisted from client null), so it's not the root cause, but preserving it via spread is still correct.

## Summary by CodeRabbit

* **New Features**
  * Enhanced transaction data model to include import and matching metadata for improved transaction tracking.

* **Bug Fixes**
  * Fixed transaction updates to properly preserve existing server-stored fields, preventing unintended data loss during modifications.


opentabs-dev (approved):
Approving — the fix is well-reasoned, the investigation trail is rigorous (UI payload capture → controlled live experiment → post-fix re-verification), and the `...existing` spread is the right long-term pattern. Any future wire fields YNAB adds will round-trip by default instead of being silently nulled. Surfacing `imported_date`, `ynab_id`, `matched_transaction_id`, and `source` on read tools is a nice follow-on for future dedup investigations.

A few minor, non-blocking suggestions:

**1. `matched_transaction_id` schema description should note the empty-string fallback.**
`mapTransaction` returns `''` for the null case, consistent with `ynab_id` / `source`, but the current describe() doesn't mention it. Consider appending something like "(empty string if unmatched)" so LLM tool-callers don't assume nullable.

**2. `create-transaction.ts` now diverges from `update-transaction.ts` in spread discipline.**
This is correct (nothing to preserve on create), but worth a one-line comment there so a future reader doesn't assume one is bit-rot. Not needed now — just flagging.

**3. No automated regression test.**
Given the blast radius (silent duplicate imports, user-visible data corruption), a unit test asserting `updatedTransaction` preserves `existing.imported_date` when present would lock this regression shut. Plugins in this repo generally don't have unit tests, so this isn't a blocker — but this fix is a good candidate for the first one.

**4. `package-lock.json` `"peer": true` on zod** is an unrelated lockfile side effect from a re-install picking up a newer peerDependency declaration. Harmless.

Nice work on the investigation — the PR body itself is documentation future-you will thank you for.

tresni:
I'm sorry, I saw that you had marked it as approved and didn't think I needed to do anything more with it. Are there changes that you need me to make to be able to merge?

PR #73, opened 2026-04-24
fix(teams): capture MSAL token via pre-script storage observer

### LINKED ISSUE #56: Microsoft Teams Plugin not working - v0.0.84
The scope patch (api.spaces.skype.com → ic3.teams.office.com) makes isReady() pass, but the authz endpoint (/api/authsvc/v1.0/authz) returns 401 because it doesn't accept the ic3.teams.office.com scoped MSAL token. 
Teams no longer stores an api.spaces.skype.com scoped token in localStorage at all — the entire Skype JWT exchange flow appears to be deprecated.
I had Claude do some exploring to try and fix but it couldn't.

Brandon-Haney:
## Summary

Fixes #56 — Microsoft Teams plugin not working on enterprise (`teams.microsoft.com`).

Microsoft migrated the Teams SPA to the MSAL v2 cache format. Cache keys are now pipe-separated (`msal.2|<accountId>|login.windows.net|accesstoken|<clientId>|<tenant>|<scopes>|`) with scopes as a space-separated list, instead of the v1 dash-separated layout with a trailing `--` sentinel. The plugin's key matchers used `endsWith(scopeSuffix)` and `'-idtoken-'`, both v1-only — so `isReady()` always returned `false` on enterprise.

This PR replaces the `endsWith`-based matchers with separator-agnostic substring checks (clientId + credential type + scope) that work for **both** MSAL v1 and v2 layouts. The scope strings drop their trailing `--` so they're plain scope values, not key-suffix fragments.

## Diagnosis vs. the original report

The reporter's theory was that the `api.spaces.skype.com` scope was removed from Teams entirely and the Skype JWT exchange was deprecated. Verified directly against a live enterprise tab that this is not the case.

Sampled the matching token on `teams.microsoft.com`:

- Key format: `msal.2|<accountId>|login.windows.net|accesstoken|<clientId>|<tenant>|https://api.spaces.skype.com/authorization.readwrite https://api.spaces.skype.com/region.readwrite https://api.spaces.skype.com/user_impersonation https://api.spaces.skype.com/.default|`
- JWT `iat`: 15 minutes before the check — MSAL is **actively acquiring** this scope right now, it isn't a stale cache.
- JWT `aud`: `https://api.spaces.skype.com`
- JWT `scp`: `Authorization.ReadWrite Region.ReadWrite user_impersonation`
- JWT `iss`: standard Azure AD STS

A 15-minute-old freshly-minted token from Azure AD is proof the scope is alive and refreshable. The fix is a matcher rewrite, not an auth-flow rewrite or a new plugin split.

## Verification

On enterprise `teams.microsoft.com`:
- `isReady()` returns `true`
- `/api/authsvc/v1.0/authz` returns `200`
- Chat tools (`list_conversations`, `read_messages`, `send_message`) succeed

Consumer `teams.live.com` continues to work — the new matcher handles both v1 and v2 formats, so the consumer path (which was already working) is unaffected.

## Follow-up (not in this PR)

The same MSAL v2 rollout likely affects other Microsoft plugins that still match cache keys by the legacy `--` suffix or `-idtoken-` fragment: `microsoft-word`, `onenote`, and `outlook`. They'll need a similar matcher update. Leaving that as a separate follow-up since I haven't reproduced the issue on those plugins.

## Test plan

- [x] Load the extension on an enterprise `teams.microsoft.com` tab and confirm the plugin reports ready in the side panel
- [x] Invoke `teams_list_conversations` / `teams_read_messages` and confirm success
- [ ] Load on consumer `teams.live.com` and confirm no regression

## Summary by CodeRabbit

* **New Features**
  * Adds a startup pre-script that captures and surfaces Teams access tokens and sign-in names for the plugin.
  * Plugin manifest updated to run the pre-script at startup.

* **Bug Fixes**
  * More reliable token detection and usage for authenticated requests, improving sign-in and session handling across consumer and enterprise scenarios.
  * More robust JWT decoding and clearer authentication error messages.


opentabs-dev:
Thanks for the careful diagnosis here — the live-token sample proving the scope is still refreshable is exactly the kind of evidence that makes this PR reviewable. You're right that lewis-chi's "scope deprecated / Skype JWT exchange dead" theory was wrong.

That said, we'd like this fix to land on the pre-script SDK we shipped yesterday rather than as a matcher rewrite. Docs: https://opentabs.dev/docs/sdk/utilities#pre-script-utilities

Reason isn't this PR specifically — your matcher fix works today and is the smallest possible diff. It's that Microsoft has changed this cache layout twice in recent memory (v1 → v2 key format is the second time), and the Microsoft-plugin cluster (teams, outlook, onenote, microsoft-word) keeps re-hitting the same class of bug from a different angle each time. Every round costs a user-reported issue, a contributor diagnosis, a review, and a release. A pre-script that captures tokens from outgoing `fetch` calls is format-agnostic — whatever Microsoft does to the cache next, the token still has to flow through `fetch` to reach the wire, and the pre-script sees it there.

Concretely for Teams enterprise:

```typescript
// plugins/teams/src/pre-script.ts

export default definePreScript(({ set }) => {
  const real = window.fetch;
    const url = typeof input === 'string' ? input : input instanceof Request ? input.url : input.toString();
    // Scope to the Skype API surface so we don't grab unrelated Graph tokens
    if (auth && /api\.spaces\.skype\.com/.test(url)) set('skypeAuth', auth);
    return real(input, init);
  };
});
```

Adapter reads with `getPreScriptValue<string>('skypeAuth')`. `plugins/prescript-test/` is a working reference.

One honest tradeoff worth calling out: pre-scripts only fire on future navigations, so on first install a user with Teams already open needs to reload once. After that, plugin updates auto-reload matching tabs (`message-router.ts` handles `preScriptHash` bumps). The matcher-only approach doesn't have that first-install hiccup, which is why this is a judgment call rather than a clear win — but we think the long-run maintenance story tips it toward pre-scripts for the Microsoft cluster specifically.

Would you be willing to rework the fix onto the pre-script SDK? Happy to review. The PascalCase body normalization and other plugin-specific work in your current diff stays exactly where it is — only the token discovery path moves.

Brandon-Haney:
Thanks for the SDK pointer — agree with the long-term direction. I rebuilt on the pre-script SDK end-to-end and pushed the rework as a single commit.

One empirical finding worth surfacing before you review: the fetch-interception pattern from your example captures nothing in Teams enterprise web v2. The Skype-scoped MSAL access token (`aud=https://api.spaces.skype.com`) is never sent through main-world `fetch` in this environment — verified across `fetch` (init.headers + `Request.headers`), `XMLHttpRequest.setRequestHeader`, `WebSocket` constructor, `navigator.sendBeacon`, and `chrome.debugger`'s Network domain (which also sees Service Worker / iframe / worker traffic). 300-request capture across refresh + Chat-tab nav + opening a thread:

- Requests to `*.spaces.skype.com`: **0**
- Requests to `*.asm.skype.com`: **0**
- 14 outbound `Authorization: Bearer` requests, all to `asyncgw.teams.microsoft.com`, `teams.microsoft.com` (registrar / trap / tokens / settings), `substrate.office.com`, `webshell.suite.office.com`, `loki.delve.office.com` — different scopes (one sampled JWT was `aud=ic3.teams.microsoft.com`)

The Skype-scoped token only ever flows MSAL → plugin, which POSTs it to `/api/authsvc/v1.0/authz`; the page itself never sends it on the wire. PR #69's Outlook-on-`cloud.microsoft` case differed on exactly this point — MSAL's Protected Token Cache encrypted localStorage there, so the page-side fetch was the only observable path.

So I kept the pre-script SDK but observed the cache layer directly instead — `Storage.prototype.setItem` + an initial `localStorage` scan, recognising entries by their **value** fields (`credentialType`, `target`, `secret`, `expiresOn`) rather than key shape. Same format-agnostic property: doesn't depend on `endsWith(suffix)` or any separator, and survives a v3/v4 key layout the same way fetch interception would have. Verified end-to-end on the live enterprise tenant — pre-script captures both `enterpriseToken` and the ID-token-derived `signInName`, `isReady()` returns true, `teams_get_current_user` returns the live identity.

One thing worth a follow-up issue: `getPreScriptValue` reads via `globalThis.__openTabs._pluginName`, but the adapter IIFE wrapper only binds `_pluginName` inside the tool-dispatch handle wrapper — not at adapter init. So `isReady()` (which runs before any tool dispatch) gets `undefined` from the SDK helper, even when the pre-script captured the value. Every plugin's `isReady()` does an auth check, so this collides with the universal pattern as soon as a plugin migrates auth to a pre-script. Worked around with a small `readPreScriptValue` shim that falls back to a direct namespace read; happy to file as a follow-up.

Brandon-Haney:
Addressed in 66971bf8:

- **base64url padding in `decodeJwtClaims` / `decodeJwtPayload`** — fixed in both the pre-script and adapter. The catch was silently masking the `atob` throw on payloads where length % 4 ≠ 0, which is the majority of JWTs.
- **PII in `log.debug`** — value dropped from the message; only logs the fact now.
- **Idempotent `setItem` patch** — marker symbol added on `Storage.prototype` so re-injection (dev hot reload, future realm reuse) doesn't stack wrappers.
- **Hostname-based slot routing** — switched to `window.location.hostname === 'teams.live.com'` to keep the pre-script's slot decision consistent with the adapter's `detectEnvironment()` by construction.

Deferring the "instrument captured-but-expired token" suggestion: that branch is hit on every `isReady()` poll between MSAL refreshes, so a naive `console.warn` there would spam. A one-shot or rate-limited warn would be useful for production debugging but isn't a correctness fix — happy to do as a follow-up issue.

Brandon-Haney:
Addressed in 537abd5d:

- **`inspect` null guard** — `JSON.parse(value)` can return `null`, primitives, or arrays. Reading `parsed.credentialType` on `null` throws, and a throw in the initial-scan loop jumps to the outer catch and aborts the scan, so a single stray entry could mask a token that comes after it. Added an `object && !null && !Array.isArray` guard before property access.
- **Docblock cleanup** — trimmed the historical line about MSAL key-format changes and the empirical verification list across fetch/XHR/WebSocket/sendBeacon/chrome.debugger. Kept the present-tense rationale for hooking `setItem`. The empirical findings still live in the PR description and the commit body.

Declining the **`readPreScriptValue` removal** suggestion — leaving the shim in place with the existing comment block, but happy to talk through the call.

The reason `readPreScriptValue` exists is that `getPreScriptValue` is unusable from `isReady()`: the SDK helper reads via `globalThis.__openTabs._pluginName`, but the adapter IIFE wrapper only binds `_pluginName` inside the tool-dispatch handle wrapper. `isReady()` runs before any tool is dispatched, so the SDK helper returns `undefined` there. I confirmed this empirically while reworking the PR — the pre-script captured the token correctly and the namespace was populated, but `isReady` reported false-not-ready until the shim was added.

I'd rather not add a TODO either — the project's own CLAUDE.md prohibits TODO/FIXME/HACK markers ("if something needs to be done, do it now"), so the shim with its present-tense rationale comment is the cleanest in-tree expression of the workaround until the SDK is fixed.

Removing the shim before the SDK fix lands would break `isReady()` for this plugin (and for every plugin that follows the universal `isAuthenticated() || waitForAuth()` pattern in `isReady()` — which is most of them — once they migrate auth capture to a pre-script). I'd be very happy to file the SDK fix as a follow-up issue against opentabs-dev/opentabs and link it from this PR if that's the preferred path; the suggested fix from earlier diagnosis was one of:

1. Bind `_pluginName` at adapter init time (so it's bound for `isReady`, lifecycle hooks, and anything else that runs before the first tool dispatch).
2. Build-time injection of the plugin name as a baked constant into the SDK reader.
3. Document that `getPreScriptValue` is only safe inside tool handlers and provide a separate API for `isReady`-time access.

Verified end-to-end on the live enterprise tenant after this push: `teams_get_current_user` still returns the expected identity.

PR #86, opened 2026-05-05
fix(teams): capture Skype JWT via authsvc fetch interceptor for Teams v2

### LINKED ISSUE #85: Teams plugin: isReady() always times out on Microsoft Teams v2 (teams.microsoft.com/v2/)
## Bug Report (filed with OpenTabs :) )

**Plugin:** `teams` v0.0.84  
**Teams URL:** `https://teams.microsoft.com/v2/`  
**OpenTabs extension:** WebSocket connected (`ws://localhost:9515/ws`)

### Problem

Adapter injects and hash matches correctly, but `isReady()` always times out. Plugin stays `unavailable` regardless of Teams view (Calendar, Chat list, open thread).

### Diagnostic

`extension_check_adapter` result:
```json
{
  "adapterPresent": true,
  "adapterHash": "2a794db96d0382df178bed337e15aab0bb7c314ae6f01fdfdcdf21b8b31c7c3b",
  "hashMatch": true,
  "isReady": false,
  "toolCount": 11
}
```

Logs show repeated: `[opentabs] isReady() timed out for plugin "teams" in tab <tabId>`

### Steps to Reproduce

1. Open `https://teams.microsoft.com/v2/` in Chrome
2. Wait for full load
3. Navigate to Chat, open a conversation
4. Run `extension_check_adapter({ plugin: "teams" })`
5. `isReady` = false every time

### Environment

- Teams v2 only (org does not allow classic Teams)
- Extension reloads + tab refreshes do not resolve the issue

### Expected

`isReady()` returns `true` once Teams chat is loaded and a conversation is visible.

### Hypothesis

Teams v2 changed the DOM/JS API that `isReady()` waits on. Adapter injects fine but the readiness signal it checks no longer exists in v2.

lewis-chi:
Fixes #85.

## Problem

Teams v2 (`teams.microsoft.com/v2/`) broke the existing auth capture approach in two ways:

**1. MSAL tokens are AES-GCM encrypted in localStorage.**

Classic Teams stores MSAL cache entries as plaintext JSON:
```json
{
  "credentialType": "AccessToken",
  "secret": "eyJ0eXAiOiJKV1Q...",
  "target": "https://spaces.skype.com/...",
  "expiresOn": "1746400000"
}
```

Teams v2 stores the same entries encrypted:
```json
{
  "id": "...",
  "nonce": "<base64>",
  "data": "<base64 AES-GCM ciphertext>"
}
```

The pre-script's `credentialType` / `secret` checks find nothing — `isTeamsAuthenticated()` returns `false`, tab stays `unavailable`.

**2. `LokiAuthToken` in sessionStorage has the wrong audience.**

Teams v2 stores a `LokiAuthToken` JWT in sessionStorage, but its `aud` claim is `394866fc-...` (the Loki service), not the Skype API. It cannot be passed to the authsvc endpoint or used for chat API calls.

## Fix

Add a `window.fetch` interceptor in the pre-script (Path 3) that hooks Teams' own authsvc call at startup:

```typescript
// Intercept Teams' own POST to /api/authsvc/v1.0/authz
window.fetch = function patchedFetch(input, init) {
  const url = input instanceof Request ? input.url : String(input);
  if (AUTHSVC_PATTERN.test(url)) {
    return realFetch(input, init).then(response => {
      if (response.ok) {
        response.clone().json().then(data => {
          // Enterprise: { tokens: { skypeToken, expiresIn } }
          // Consumer:   { skypeToken: { skypetoken, expiresIn } }
          set('skypeJwt', { secret: jwt, expiresOn: ... });
        });
      }
      return response;
    });
  }
  return realFetch(input, init);
};
```

The interceptor runs at `document_start` in MAIN world — guaranteed to be in place before Teams JS makes its startup authsvc call. It fires on every token refresh, keeping the captured JWT fresh without polling.

The adapter reads `skypeJwt` first in `getSkypeJwt()`:

```typescript
const preCapture = readPreScriptValue<{ secret: string; expiresOn: number }>('skypeJwt');
if (preCapture?.secret && preCapture.expiresOn > Date.now() / 1000 + 60) {
  cachedSkypeJwt = { token: preCapture.secret, expiresAt: preCapture.expiresOn * 1000 };
  return preCapture.secret;
}
// Classic Teams fallback: exchange MSAL token via authsvc ourselves
```

`LokiAuthToken` is retained as a readiness signal only — `isTeamsAuthenticated()` now accepts either the classic MSAL Skype token or the Loki token, so the tab reaches `ready` state on Teams v2.

## Auth paths summary

| Path | Trigger | Slot | Used for |
|------|---------|------|----------|
| MSAL localStorage observer | `setItem` on localStorage | `enterpriseToken` / `consumerToken` | Classic Teams (plaintext MSAL) |
| Loki sessionStorage observer | `setItem` on sessionStorage | `lokiToken` | Teams v2 readiness signal only |
| authsvc fetch interceptor | Teams' own authsvc response | `skypeJwt` | Teams v2 primary auth (encrypted MSAL) |

## Test plan

- [x] `extension_check_adapter` returns `isReady: true` on `teams.microsoft.com/v2/` after tab load
- [x] `teams_list_conversations` returns conversations
- [x] `teams_send_message` successfully sends a message
- [x] Classic Teams (`teams.microsoft.com` without `/v2/`) still works via MSAL localStorage path

## Release Notes

* **New Features**
  * Extended Teams authentication to capture Teams v2 credentials through enhanced credential detection
  * Implemented automatic fallback mechanism that reverts to existing authentication when Teams v2 credentials are unavailable or expired
  * Improved authentication state detection to recognize both legacy and Teams v2 tokens for seamless version compatibility


plugins/teams/src/pre-script.ts
```diff
@@ -95,7 +107,101 @@ definePreScript(({ set, log }) => {
...
+
+  /**
+   * Teams v2 (teams.microsoft.com/v2/) stores a Loki bearer token as a raw JWT
+   * string in sessionStorage under the key `LokiAuthToken`. This token is used
+   * only as a readiness signal — its audience is the Loki service, not the
+   * Skype API, so it cannot be used directly for authsvc or chat APIs. The Skype
+   * JWT is captured separately via the authsvc fetch interceptor (Path 3).
+   */
+  const inspectLokiEntry = (key: string, value: string): void => {
+    if (!key.startsWith('LokiAuthToken') || key.startsWith('EXPLoki')) return;
```
opentabs-dev, line 122: This guard is unreachable: if `key.startsWith('LokiAuthToken')` is true, `key.startsWith('EXPLoki')` is necessarily false. If the intent is to skip an `EXPLokiAuthToken`-style variant, the condition is inverted — the check probably should be `if (!key.startsWith('LokiAuthToken') || key.startsWith('EXPLokiAuthToken')) return;` or similar. Either way, please either fix the intent or drop the dead clause.

plugins/teams/src/pre-script.ts
```diff
@@ -95,7 +107,101 @@ definePreScript(({ set, log }) => {
...
+              .then((data: unknown) => {
+                if (!data || typeof data !== 'object') return;
+                const d = data as Record<string, unknown>;
+
+                // Enterprise format: { tokens: { skypeToken: string, expiresIn: number } }
+                const tokens = d.tokens as Record<string, unknown> | undefined;
+                const entJwt = typeof tokens?.skypeToken === 'string' ? tokens.skypeToken : null;
+                if (entJwt) {
+                  const expiresIn = typeof tokens?.expiresIn === 'number' ? tokens.expiresIn : 3600;
+                  set('skypeJwt', { secret: entJwt, expiresOn: Math.floor(Date.now() / 1000) + expiresIn } as CapturedToken);
```
opentabs-dev: Biome `format:check` wants this multi-line — CI fails here. `npm run format` in `plugins/teams/` will fix it (same for line 184).

plugins/teams/src/pre-script.ts
```diff
@@ -95,7 +107,101 @@ definePreScript(({ set, log }) => {
...
+   */
+  const FETCH_MARKER = '__opentabsTeamsFetchPatched';
+  const winAny = window as unknown as Record<string, unknown>;
+  if (!winAny[FETCH_MARKER]) {
+    const realFetch = window.fetch.bind(window);
+    window.fetch = function patchedFetch(
+      input: Parameters<typeof fetch>[0],
+      init?: Parameters<typeof fetch>[1],
+    ): ReturnType<typeof fetch> {
+      const url = input instanceof Request ? input.url : String(input);
```
opentabs-dev, line 158: Minor: misses the `URL` object case (valid `fetch` input). `String(urlObj)` happens to return `href`, so it works, but `input instanceof Request ? input.url : input instanceof URL ? input.href : String(input)` is clearer about intent.

plugins/teams/src/pre-script.ts
```diff
@@ -95,7 +107,101 @@ definePreScript(({ set, log }) => {
...
+              .catch(() => {
+                // Never block the page on JSON parse failure.
+              });
+          }
+          return response;
+        });
+      }
+      return realFetch(input, init);
+    };
+    winAny[FETCH_MARKER] = true;
```
opentabs-dev, line 203: The `setItem` patch below stamps its marker on `Storage.prototype`, so it's idempotent even if the adapter is re-injected in the same realm. This fetch patch stores the marker on `window` but replaces `window.fetch` — if the adapter is re-injected into the same realm (the dev hot-reload case the setItem comment calls out), the second injection captures the *already-patched* fetch as `realFetch` and stacks wrappers.

Safer: stamp the marker on the wrapper function itself (e.g., `patchedFetch.__opentabsPatched = true`) and skip if `window.fetch.__opentabsPatched`. That makes the guard robust to fresh `window` objects that still carry the patched fetch.

opentabs-dev (commented):
Solid approach — intercepting the authsvc response at document_start is the right call, and the three-path design is well-organized. A few things to address before merge:

1. **`format:check` fails** on `plugins/teams/src/pre-script.ts` — the two `set('skypeJwt', {...})` calls need to be wrapped multi-line. Run `npm run format` in `plugins/teams/`.
2. **`getSkypeIdentity()` still hard-requires the MSAL localStorage token** (teams-api.ts around line 254, outside the diff) — that defeats the PR's goal on Teams v2 where MSAL is encrypted. It should mirror `getSkypeJwt()`'s priority: prefer the pre-captured `skypeJwt` (decode its claims for `skypeid`; read `signInName` from the pre-script), and only fall back to the authsvc exchange on classic Teams. Any tool that resolves an MRI through `getSkypeIdentity()` will throw on v2 today. The PR body claims `teams_send_message` works on v2 — worth confirming which tools actually hit `getSkypeIdentity()` there.
3. **Contradictory key guard** in `inspectLokiEntry` — see inline.

Also worth addressing: the fetch patch is per-window whereas the setItem patch is per-prototype — the former can stack wrappers on re-injection. Details inline.

PR #100, opened 2026-05-28
feat: SharePoint/OneDrive support for Excel, Word, and PowerPoint plugins

Brandon-Haney:
# feat: SharePoint/OneDrive support for Excel, Word, and PowerPoint plugins

## Summary

The Excel Online, Microsoft Word, and PowerPoint plugins currently activate only on the standalone `*.cloud.microsoft` apps. The vast majority of Office files in enterprise tenants are opened through SharePoint or OneDrive (`*.sharepoint.com/:x:/...`, `:w:`, `:p:`), where the plugins do nothing today. This PR adds SharePoint/OneDrive support to all three plugins behind a single shared pattern.

## Why it needs a different auth path

On the standalone `excel.cloud.microsoft` / `word.cloud.microsoft` / `powerpoint.cloud.microsoft` apps, MSAL writes plaintext access-token entries into `localStorage` keyed by client id and scope — the plugins already read those. SharePoint-hosted Office, by contrast, keeps the MSAL cache **encrypted** in `localStorage` (the SharePoint shell stores it under different keys and protects the token material), and the top frame never calls the Microsoft Graph workbook/document/presentation API directly — the editor runs in a cross-origin WAC frame. The existing plaintext-localStorage read finds nothing.

What the SharePoint top frame **does** do on load is mint per-resource Graph tokens by POSTing to `https://login.microsoftonline.com/<tenant>/oauth2/v2.0/token`. The response is a plaintext JSON body containing `access_token`, `scope`, and `expires_in`.

## Approach

A pre-script (`definePreScript`) wraps `window.fetch` in the MAIN world at `document_start` and inspects:

1. **AAD token-endpoint responses** (`/oauth2/v2.0/token`) — when the response grants a Graph scope, stash the token + expiry.
2. **Direct Graph requests** (`graph.microsoft.com`) — as a fallback, capture the `Authorization: Bearer <token>` header on outbound Graph calls.

Captured tokens are stashed both in the pre-script's per-load namespace and mirrored to `localStorage` under a per-plugin key. The mirror is the only way warm reloads and same-origin sibling tabs see a token without waiting for MSAL to mint a fresh one.

Tools then read the captured token first; the existing plaintext-MSAL-localStorage read is the fallback. For Word and PowerPoint, the workbook/document/presentation context (driveId + itemId) comes from URL query params first, then from `_wopiContextJson.DriveId` (PowerPoint), then from a Graph `/shares` lookup using the sharing URL as a base64url share id.

### Three commits, one shared pattern

* `feat(excel-online): support SharePoint/OneDrive-hosted workbooks`
* `feat(microsoft-word): support SharePoint/OneDrive-hosted documents`
* `feat(powerpoint): support SharePoint/OneDrive-hosted presentations`

The pre-scripts are intentionally three near-identical files rather than one shared module — each plugin builds independently with its own SDK version, so a shared dep would mean another package and another version to coordinate. Names of `localStorage` keys, `Symbol.for` markers, and log prefixes are per-plugin to keep two plugins on the same origin from colliding.

The PowerPoint commit additionally turns `getCurrentDriveId()` from sync to async (the Graph `/shares` resolution is async). All 19 existing file/folder/slide tools that called it get an `await` added — purely mechanical.

## Hardening already applied (informed by CodeRabbit on PR #73)

PR #73's review surfaced a set of patterns worth getting right the first time. All three plugins now follow them:

* **Idempotent fetch patch.** Each pre-script's wrapper is marked with `Symbol.for('opentabs.<plugin>.fetch.patched')`. A second injection into the same realm (hot reload, future iframe-reuse) returns early instead of stacking wrappers and recursing.
* **URL host checks parse the URL.** `new URL(url).hostname.toLowerCase() === 'graph.microsoft.com'` (and same for `login.microsoftonline.com`) replaces boundary-aware-but-still-substring regex. Avoids the CodeQL "Incomplete URL substring sanitization" alert.
* **AAD scope-claim check parses each scope.** The `scope` claim is space-separated; we split and parse each entry as a URL, matching by hostname rather than substring-matching the whole claim.
* **MSAL fallback requires strict expiration.** `expiresOn > now` — a missing or unparseable `expiresOn` is treated as expired rather than usable. Matches the fix CodeRabbit required on PR #73's `findMsalToken`.

Three PowerPoint-specific bugs were caught and fixed in the same pass:

* `isSharePoint()` was `/\.sharepoint\.com/i.test(getCurrentUrl())` — a substring match over the full URL, the same CodeQL pattern. Now parses with `new URL()` and checks `hostname.endsWith('.sharepoint.com')`, matching the Excel and Word implementations.
* The MSAL active-account fallback in `getDriveIdSync` had `match[1]?.replace('-', '')` — only the first hyphen is replaced. The MSAL account id contains multiple hyphens; uses `/-/g` now.
* `getMsalToken()` didn't filter by `MSAL_CLIENT_ID`, so a Graph access-token entry from another Microsoft 365 app sharing the page realm could be returned. Now matches the Excel and Word implementations.

Excel also had a quiet bug: `Number.parseInt(parsed.expires_on, 10)` checked a snake_case key that MSAL doesn't write — Word and PowerPoint correctly use `expiresOn`. The Excel check never matched, and stale tokens were silently returned. Fixed.

## Verification

Each plugin's `npm run check` passes (`build + type-check + lint + format:check`).

End-to-end verified on a live enterprise tenant:

* **Excel**: opened a `*.sharepoint.com/:x:/r/personal/.../Doc.xlsx` workbook; `get_workbook_info`, `get_range`, `list_worksheets`, `update_range` all succeed.
* **Word**: opened a `*.sharepoint.com/:w:/r/sites/.../Doc.docx`; `get_active_document`, `get_document_text`, `update_document`, `append_to_document` all succeed.
* **PowerPoint**: opened a `*.sharepoint.com/:p:/r/sites/.../Deck.pptx`; `get_slides` downloaded and parsed all slides, `get_slide_content` and `get_slide_notes` work.

The standalone `excel.cloud.microsoft` / `word.cloud.microsoft` / `powerpoint.cloud.microsoft` paths continue to work — the captured-token path runs first, and the existing plaintext-MSAL-localStorage path remains as a fallback for warm tabs.

## Notes for reviewer

* The PowerPoint commit is the largest because of the `getCurrentDriveId()` sync→async ripple. Reviewing the powerpoint-api.ts + pre-script.ts diff first (where the actual logic lives) and then skimming the 19 mechanical `await` additions is the lowest-overhead path.
* A separate stacked PR (`feat/powerpoint-editing-tools`) adds 12 net-new editing tools (shapes, text boxes, images, slide duplication, layout queries, and an explicit session model). That's a distinct feature; submitting separately.

## New Features
* Extended Excel Online, Word, and PowerPoint plugins to support SharePoint and OneDrive-hosted documents alongside cloud versions.
* Added reauthentication tool for SharePoint/OneDrive scenarios to clear stale authentication state.

## Improvements
* Enhanced authentication token handling with improved capture and validation mechanisms.
* Refined plugin readiness detection for better activation on SharePoint/OneDrive documents.
* Improved error messaging for authentication failures.


Brandon-Haney:
Addressed in 50a35174.

**CodeRabbit findings — all three actionables**

1. **`unescape()` deprecation in `encodeShareId`** (Excel + Word) — replaced with `TextEncoder` + `String.fromCharCode` loop. PowerPoint had the same pattern in `powerpoint-api.ts`; fixed there too for consistency.

2. **PowerPoint `cachedDriveId` global per-tab** — the Office apps are SPAs, so same-tab navigation to a different presentation changes `getCurrentUrl()` without reloading the adapter, and the single-slot cache would silently return the wrong drive. Excel `cachedContext` and Word `cachedContext` had the same shape with the same concern; CodeRabbit only flagged PowerPoint but all three now key the cache by `getCurrentUrl()` and invalidate on URL change.

**CI failures — confirmed pre-existing on `upstream/main`, not introduced by this PR**

- `label` — `actions/labeler@v6` returns `Resource not accessible by integration` because external-fork PRs don't get the `pull-requests: write` token. Same failure on every external-fork PR.
- `plugins` — `retool` plugin fails biome lint on `upstream/main` itself (`ToolError` unused import in `tools/delete-app.ts:1`, `metadataArray` unused var in `tools/run-grpc.ts:33`). Same failure on the latest `main` CI run for `3e27b034`.
- `macos` — config-watcher, browser-permission, onboarding, side-panel-live-update specs all fail. Same failures on `upstream/main`'s macOS run for `3e27b034`. None of these tests touch the Office plugins.

The substantive checks all pass: `check` (build + lint + test), `check-docs`, `analyze` (CodeQL), `windows`, `windows-e2e`. The Linux `e2e` was still in progress when I started writing this; should be green by the time you read it.

Brandon-Haney:
Hi @opentabs-dev — gentle nudge on this PR (and its siblings #101 / #104). They're Microsoft-plugin features — Outlook calendar, MSAL v3 auth, and SharePoint/OneDrive support for the Office plugins — all CodeRabbit-clean with every finding addressed, and verified live against a real tenant.

The failing checks are pre-existing/infra, not introduced by these PRs: `label` fails on all fork PRs (the read-only token can't write labels), `plugins` is red on `retool`'s formatting on `main`, and `e2e`/`macos` are platform-test flakiness that was green when my #73 merged in April. Happy to rebase if it helps. No urgency — just flagging they're ready whenever you have a window. Thanks for building this!

opentabs-dev:
Thanks for this — the Office/SharePoint work is solid, but I'm holding the merge on the **platform change** in `platform/browser-extension/src/pre-script-registration.ts` (`allFrames: false → true`). A few things to address:

**1. The stated rationale doesn't match the code.** The comment says *"All shipped pre-scripts use a `Symbol.for(...)` marker so a second injection into the same realm is a no-op."* That isn't true today — neither shipping pre-script (`teams`, `prescript-test`) nor the SDK's `definePreScript` has such a marker, and the three new Office pre-scripts in this PR don't add one either. The change *is* actually safe, but for a different reason worth stating correctly: `allFrames: true` injects a **separate, realm-isolated execution per same-origin frame** — top-frame behavior is unchanged, and each frame's `globalThis.__openTabs` store is independent, so there's no shared state to double-write. Please fix the comment to reflect the real reason.

**2. This is a global injection-semantics change** affecting *every* plugin's pre-script, not just Office. Bundling it inside a 3-plugin feature PR couples an isolated platform decision to a feature. Please **split it into its own commit/PR** so it can be reviewed on its own merits.

**3. Cross-frame value bridging.** The feature only works if pre-script values `set()` in a same-origin sub-frame actually reach the adapter (which runs in the top frame and reads *its* `globalThis.__openTabs` via `getPreScriptValue`). Since each frame is a separate realm, please confirm/show how a token captured only in a sub-frame bridges to the top-frame adapter — otherwise the SharePoint path may silently capture nothing.

Once the platform change is split out with corrected reasoning and the bridging is confirmed, rebase onto current `main` (CI here is stale) and I'll re-review. The per-plugin Office code itself looks good.

Brandon-Haney:
Thanks for the careful read — all three are addressed.

**1 + 2. Platform change split out, with corrected reasoning → #109.**

The `allFrames: false → true` change now lives in its own PR (#109) so the injection-semantics decision is reviewed on its own merits. This PR has been rebased onto current `main` and no longer touches `pre-script-registration.ts`.

You're right that the old comment was wrong — `teams`/`prescript-test` carry no `Symbol.for(...)` marker, and `definePreScript` doesn't add one, so "all shipped pre-scripts are idempotent" was never the reason this is safe. #109's comment now states the real reason: **realm isolation**. `allFrames: true` runs a *separate* execution per same-origin frame, top-frame behavior is unchanged, and each frame's `globalThis.__openTabs` store is its own, so there's no shared in-page state to double-write.

**3. Cross-frame bridging — confirmed, and it's exactly the concern you raised.**

You're correct that the per-realm `__openTabs` namespace does **not** cross frames — a value `set()` in a sub-frame is invisible to the top-frame adapter's `getPreScriptValue`. That path alone would indeed capture nothing. The bridge is a **same-origin `localStorage` mirror**:

- The pre-script's `stash()` writes the captured Graph token to **both** the realm-local namespace **and** `localStorage` (`pre-script.ts` → `localStorage.setItem(LS_TOKEN_KEY, …)`).
- `localStorage` is shared across all same-origin frames, so the sub-frame's write is visible in the top frame.
- The adapter's `getCapturedToken()` reads the realm namespace first, then **falls back to the `localStorage` mirror** (`excel-api.ts`), so a token minted in a same-origin silent-renewal sub-frame reaches the top-frame adapter.

So `allFrames: true` is **necessary** (each frame has its own `window.fetch`/`XMLHttpRequest`; the top frame can't intercept a sub-frame's token request) and **sufficient** (the captured value rides up to the adapter through shared `localStorage`). This is why the SharePoint path captures live tokens against a real tenant rather than silently getting nothing.

Once #109 lands I'll confirm this PR is green on top of it. Thanks again!

PR #104, opened 2026-06-02
feat(outlook): add calendar support (events, scheduling, free/busy)

Brandon-Haney:
## Summary

Adds calendar support to the Outlook plugin — 9 new tools (15 → 24 total) for reading and managing calendar events, plus viewing other people's availability. Calendar data is read directly from the user's existing session token on the mail page; no need to open the Calendar UI.

### New tools

| Tool | Type | Description |
|---|---|---|
| `list_calendars` | Read | List calendars, including shared ones |
| `list_events` | Read | List events in a calendar |
| `get_calendar_view` | Read | Events in a date range, with recurring series expanded |
| `get_event` | Read | Full event details (attendees, body, reminders) |
| `get_schedule` | Read | Free/busy availability for other people |
| `create_event` | Write | Create an event or meeting |
| `update_event` | Write | Update an event |
| `delete_event` | Write | Delete, or cancel-with-notice for organizers |
| `respond_to_event` | Write | Accept / decline / tentatively accept an invite |

### Supporting auth-layer changes (`outlook-api.ts`)

These came out of debugging against a live enterprise tenant and also harden the existing mail tools:

- **Scope-aware token resolution.** Mail and calendar each resolve their own capability-scoped token, with separate auth caches, so calendar calls never inherit a mail-only token (enterprise Graph tokens often lack `Calendars.*`).
- **MSAL v3 support.** `outlook.cloud.microsoft` now stores tokens under flat `msal.3|...|accesstoken|...` keys with no index, which the v2 index-walk could not find. Added `findMsalV3Token` (scans localStorage directly), tried ahead of the v2/v1 fallbacks. This also restores mail auth on the new cache schema.
- **PascalCase request bodies on the Outlook REST API.** REST v2.0 deserializes bodies case-sensitively and requires PascalCase for both entities and OData action parameters, while Graph uses camelCase. Bodies are authored in camelCase and transformed when targeting the REST base.

## Testing

Verified live against a real Microsoft 365 mailbox through the full MCP dispatch path:

- ✅ `list_calendars` (incl. shared calendars), `list_events`, `get_calendar_view` (recurrence expansion + time-zone conversion), `get_event`
- ✅ `get_schedule` for a colleague (free/busy + working hours)
- ✅ `create_event` → `update_event` → `delete_event` round-trip (test event cleaned up)
- ✅ Mail tools regression-checked (`get_current_user`, `list_messages`)
- `respond_to_event` is code-verified only — it shares the validated action-body path of `get_schedule`, but was not exercised live to avoid sending a real response to a meeting organizer.

`npm run check` passes for the plugin (build, type-check, lint, format).

## Summary by CodeRabbit

* **New Features**
  * Full calendar support: create, update, delete, and respond to events.
  * Calendar browsing: list calendars, list events, calendar-view queries, and get event details.
  * Scheduling tools: free/busy lookups to check attendee availability and retrieve per-person schedules.
  * Timezone-aware queries and improved event fields (reminders, attendees, recurrence, online meetings).

* **Documentation**
  * Updated plugin description and docs to include expanded calendar capabilities and revised Tools overview.

### DIFF
--- a/plugins/outlook/README.md
+++ b/plugins/outlook/README.md
@@ -19,7 +19,7 @@ npm install -g @opentabs-dev/opentabs-plugin-outlook
 1. Open [outlook.cloud.microsoft](https://outlook.cloud.microsoft) in Chrome and log in
 2. Open the OpenTabs side panel — the Microsoft Outlook plugin should appear as **ready**
 
-## Tools (15)
+## Tools (24)
 
 ### Account (1)
 
@@ -50,6 +50,20 @@ npm install -g @opentabs-dev/opentabs-plugin-outlook
 | Tool | Description | Type |
 |---|---|---|
 | `list_folders` | List mail folders | Read |
+
+### Calendar (9)
+
+| Tool | Description | Type |
+|---|---|---|
+| `list_calendars` | List calendars, including shared ones | Read |
+| `list_events` | List events in a calendar | Read |
+| `get_calendar_view` | View events in a date range (expands recurrences) | Read |
+| `get_event` | Get full event details | Read |
+| `get_schedule` | View others' free/busy availability | Read |
+| `create_event` | Create an event or meeting | Write |
+| `update_event` | Update an event | Write |
+| `delete_event` | Delete or cancel an event | Write |
+| `respond_to_event` | Accept, decline, or tentatively accept an invitation | Write |
 
 ## How It Works
 
--- a/plugins/outlook/package.json
+++ b/plugins/outlook/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@opentabs-dev/opentabs-plugin-outlook",
-  "description": "OpenTabs plugin for Microsoft Outlook",
+  "description": "OpenTabs plugin for Microsoft Outlook — email and calendar",
   "publishConfig": {
     "access": "public"
   },
@@ -18,7 +18,7 @@
   ],
   "opentabs": {
     "displayName": "Microsoft Outlook",
-    "description": "OpenTabs plugin for Microsoft Outlook",
+    "description": "OpenTabs plugin for Microsoft Outlook — email and calendar",
     "urlPatterns": [
       "*://outlook.cloud.microsoft/*",
       "*://outlook.live.com/*",
--- a/plugins/outlook/src/index.ts
+++ b/plugins/outlook/src/index.ts
@@ -2,24 +2,34 @@ import { OpenTabsPlugin } from '@opentabs-dev/plugin-sdk';
 import type { ToolDefinition } from '@opentabs-dev/plugin-sdk';
 import { isAuthenticated, waitForAuth } from './outlook-api.js';
 import { createDraft } from './tools/create-draft.js';
+import { createEvent } from './tools/create-event.js';
+import { deleteEvent } from './tools/delete-event.js';
 import { deleteMessage } from './tools/delete-message.js';
 import { forwardMessage } from './tools/forward-message.js';
 import { downloadAttachment } from './tools/download-attachment.js';
 import { getAttachmentContent } from './tools/get-attachment-content.js';
+import { getCalendarView } from './tools/get-calendar-view.js';
 import { getCurrentUser } from './tools/get-current-user.js';
+import { getEvent } from './tools/get-event.js';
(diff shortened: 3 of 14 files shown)


Brandon-Haney:
Thanks for the review — addressed in 4be261ff, all verified live against a real mailbox.

✅ **Empty/202 responses** — `sendRequest` now gates JSON parsing on the response content-type, so action endpoints returning 202/empty bodies (cancel, RSVP) and `sendMail` return success instead of throwing on `.json()`.

✅ **Calendar read vs write scope** — added a dedicated `calendar-write` capability requiring `Calendars.ReadWrite`. `create_event`, `update_event`, `delete_event`, and `respond_to_event` resolve through it, so a read-only token can never bind to a mutating call.

✅ **Path-encode IDs** — `encodeURIComponent` applied to every opaque event/calendar ID across the calendar tools.

✅ **No-op `update_event`** — enforced as a handler-level guard that throws a validation error. Worth noting for future reference: a Zod `.refine()` does **not** survive the JSON-Schema conversion the platform uses for MCP input validation (confirmed live — a refine-based guard never fired), so the check has to live in the handler.

✅ **`get_calendar_view` time zone** — corrected. The range bounds derive from the offset in the start/end timestamps (UTC when none is given); `time_zone` only controls the zone of the returned event times. Tool/param descriptions updated to match.

⏭️ **Handler signature `(params, context?)`** — skipped intentionally. `context` is an optional second argument used only for progress reporting; all 15 existing Outlook tools use the single-parameter `async params =>` form and these tools don't report progress, so this matches the established convention rather than diverging from it.

opentabs-dev:
Thanks for this! Heads up: #101 (MSAL v3 + token cascade) just merged to `main` and it **rewrote `outlook-api.ts`'s auth model**, which conflicts with this PR.

The two PRs took different approaches to the same problem:
- **#101 (now on main):** removed scope pre-filtering — it collects *all* candidate tokens and cascades through them on 401/403, caching the first the API accepts. Added MSAL v3 support + security hardening (URL-hostname scope matching, strict `expiresOn`).
- **This PR:** introduces a `Capability` (mail/calendar) concept with per-capability scope pre-filtering and separate auth caches (`outlook` / `outlook-calendar`), built on the pre-#101 code.

Because #101's cascade already tries every token regardless of scope, the per-capability pre-filter here is largely redundant now — though you may still want the **separate calendar cache** so calendar calls don't thrash a mail-only token. Since that's a design judgment on your auth flow, could you rebase `outlook-api.ts` onto the new cascade model? The calendar **tools** themselves (`create-event`, `get-calendar-view`, etc.) are clean and conflict-free — it's only the shared API file that needs reconciling.

Happy to help if you'd like me to take a pass at the rebase instead — just let me know.

Brandon-Haney:
Thanks @opentabs-dev — rebased onto the cascade. `outlook-api.ts` is now #101's model plus two additions, and the calendar tools are unchanged. PR shows mergeable again.

Reconciliation:

- **Dropped the per-capability scope pre-filter.** You're right that the cascade makes it redundant — it discovers the working token empirically on 401/403, so pre-filtering by `Calendars.*` only risked discarding a token the API would actually accept.
- **Kept a separate calendar cache** (`outlook` / `outlook-calendar` / `outlook-calendar-write`). The cascade caches a single winner; with mail and calendar sharing one slot they evict each other on every alternation — a mail call caches a mail token, the next calendar call 403s and re-cascades to a calendar token, then the next mail call 403s on *that*. A per-capability cache key gives each its own slot so each pays the cascade only once. `capability` now **only** selects the slot — it does not pre-filter candidates.
- **Carried over one thing the cascade lacked:** a `pascalCaseKeys` transform on request bodies for the Outlook REST base. REST deserializes case-sensitively, so event writes (`Subject`, `Start`/`DateTime`) and OData action params (`Schedules`, `Comment`) 400 without it. Graph (mail) is camelCase and unaffected. This is independent of calendar — might be worth pulling into the mail path on `main` too, since `reply`/`forward` against a REST-only tenant hit the same wall.

Net `outlook-api.ts` diff vs main is +69 lines on top of the cascade rather than a competing rewrite. Build / type-check / lint / format all pass. Happy to split the `pascalCaseKeys` bit into its own PR if you'd prefer to keep this one calendar-only.

PR #116, opened 2026-06-17
fix(office): capture refreshed Graph token via XHR on SharePoint

Brandon-Haney:
## Problem

On a long-lived SharePoint-hosted Office tab, the **PowerPoint** and **Word** plugins fail every tool call with `AUTH_ERROR — Not authenticated` after roughly an hour, even though the adapter is injected and ready.

Root cause: both pre-scripts only wrapped `window.fetch`. They capture the Graph token minted on a **cold page load**, but SharePoint's WAC/Owl framework performs its subsequent **silent token refreshes over `XMLHttpRequest`**, not `fetch`. So the captured token (and its `localStorage` mirror) is never refreshed — it expires after the first hour and the adapter correctly rejects the stale token, leaving the plugin unauthenticated until the page is manually reloaded.

This was diagnosed live: the pre-script namespace held a token with `exp` ~40 minutes in the past.

## Fix

Port the interceptor pattern already proven in `excel-online`:

- **Wrap `XMLHttpRequest` in addition to `fetch`**, so Graph-scoped tokens are captured from AAD token-endpoint responses for the tab's entire lifetime — including the hourly silent refresh.
- **Broaden the token-endpoint matcher** to accept the v1 path (`/oauth2/token`) alongside v2 (`/oauth2/v2.0/token`).

Both patches are idempotent under re-injection (separate `Symbol.for` markers for fetch and XHR). No behavior change on the standalone `*.cloud.microsoft` apps.

## Verification

Verified end-to-end on a live SharePoint-hosted presentation:

- Before: captured token expired (`exp` in the past) → `get_current_user` returned `AUTH_ERROR`.
- After rebuild + reload: both `fetchPatched` and `xhrPatched` active, fresh token captured (~87 min validity), `get_current_user` and `get_slides` (14 slides) both succeed.
- `npm run build`, lint, and format:check pass for both plugins.

## Release Notes

* **Bug Fixes**
  * Enhanced Microsoft Word and PowerPoint plugins to capture and persist authentication tokens more reliably across both modern and legacy Azure AD token endpoint formats.
  * Improved interception coverage for both network fetch requests and XMLHttpRequest traffic, including direct Graph bearer requests and token-endpoint JSON responses.
  * Token handling now better validates Graph-scoped tokens and computes expiration more robustly when values are missing or malformed.
* **Documentation**
  * Updated pre-script documentation to reflect the new, format-agnostic token capture behavior.

### DIFF
--- a/plugins/microsoft-word/src/pre-script.ts
+++ b/plugins/microsoft-word/src/pre-script.ts
@@ -9,15 +9,15 @@ import { definePreScript } from '@opentabs-dev/plugin-sdk/pre-script';
  * edits through the cross-origin WOPI canvas and MSAL stores its token cache
  * encrypted, so there is no plaintext Graph token to read from `localStorage`.
  * What the page does on load is mint per-resource access tokens by POSTing to
- * the AAD token endpoint (`login.microsoftonline.com/<tenant>/oauth2/v2.0/token`)
- * via `fetch`; each response is plaintext JSON with `access_token`, `scope`, and
- * `expires_in`.
+ * the AAD token endpoint (`login.microsoftonline.com/<tenant>/oauth2/v2.0/token`);
+ * each response is plaintext JSON with `access_token`, `scope`, and `expires_in`.
  *
- * This wraps `window.fetch`, captures the Graph-scoped token from those
- * token-endpoint responses, and stashes it for the adapter to read via
- * `getPreScriptValue`. It also captures a `Bearer` token from any direct
- * `graph.microsoft.com` request, covering the standalone `word.cloud.microsoft`
- * app.
+ * This wraps both `window.fetch` and `XMLHttpRequest`, captures the Graph-scoped
+ * token from those token-endpoint responses, and stashes it for the adapter to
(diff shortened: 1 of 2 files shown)


Brandon-Haney:
Thanks for the review. Addressed and deferred:

**`{ once: true }` on the XHR `load` listener — addressed** (db7e1d17). XHR instances are reusable and we add the listener on every `open()`, so without `once` a reused instance accumulates a listener per request and re-runs capture on each later response. Scoped each listener to a single fire. The reference `excel-online` pre-script had the identical latent leak, so the same fix was applied there as well to keep the three Office pre-scripts consistent.

**Extract shared interceptor into a common module — deferred** to #118. The three Office pre-scripts are ~95% identical and the duplication is a real drift risk, but plugins are standalone packages that can't cross-import a local shared module — the only valid home is a factory exported from `@opentabs-dev/plugin-sdk`, which means an SDK API addition + publish + bumping all three plugins. That's a deliberate refactor rather than something to fold into this focused auth fix, so it's tracked separately.

PR #137, opened 2026-07-06
feat(outlook): match OWA compose — threaded drafts, signature, and default font

Brandon-Haney:
## What

Brings plugin-composed Outlook mail in line with what OWA's compose surface produces:

1. **Threaded reply/forward drafts** — `reply_to_message` / `forward_message` gain a `draft` mode that saves a threaded draft (quoted conversation, reply headers, recipients pre-filled) to Drafts for manual review.
2. **Automatic signature** — the user's Outlook signature (with its inline logo) is applied to drafts and sends.
3. **Default compose font** — plain-text bodies are wrapped in the user's default compose font (family/size/color), matching OWA.

## Why

Signatures and the default compose font are client-side OWA *compose* behaviors — Microsoft Graph's `mailboxSettings` exposes neither. The plugin's tools go straight to the API, so nothing signed or styled the body and agents had to fabricate signatures. This reads the real values from the first-party OWS gateway and applies them deterministically.

## How

- **`owsRequest()` transport** (`outlook-api.ts`): a same-origin fetch to the OWS gateway on `outlook.cloud.microsoft`, reusing the existing MSAL token cascade in its own cache slot, with the mailbox anchor derived from each token's `puid`/`tid` claims. Signature bodies live in the OWS *large-settings* collection and require the `x-islargesetting` header.
- **`compose-defaults.ts`**: resolves signature names + default font (session-cached), fetches a signature's HTML (logo inlined as a `data:` URI — self-contained, no image hosting), and composes an HTML body. **Degrades gracefully** — if settings can't be fetched, a valid plain body is still returned; a draft is never blocked.
- Wired into `create_draft`, `send_message`, `reply_to_message`, `forward_message` for both draft and immediate send. Reply/forward run both modes through `createReply`/`createForward` → PATCH → send, so font, signature, and the quoted thread are identical either way.

## Behavior changes (called out explicitly)

- `create_draft` and `send_message` now send text bodies as `contentType: HTML` so the default font can be applied — escaped text renders identically.
- Immediate reply/forward are reimplemented as `createReply`/`createForward` → PATCH → send (previously single `/reply` / `/forward` actions). A composing failure cleans up the created draft; a send failure leaves the draft in place (ambiguous — the message may already be on its way, so it is not deleted).
- New `include_signature` input (default `true`) on the four tools.

## Testing

Verified live against enterprise M365 mailbox: `create_draft`, `send_message`, and both immediate and draft reply/forward produce the default-font wrapper plus the correct signature (new-mail vs reply), with the quoted thread preserved on reply/forward. The design is account-neutral (cross-tenant token cascade; anchor derived from each token's own claims; OWA-standard setting names). Not yet verified against personal outlook.com accounts, or mailboxes whose signatures live only in the legacy single-signature store — both fall back to no signature rather than erroring.

## Summary by CodeRabbit

* **New Features**
  * Outlook email composition for drafts, replies, forwards, and sends now preserves Outlook compose formatting and automatically applies the default compose font and signature.
  * Added an option to omit the signature.
  * Replies and forwards can be saved as drafts; when enabled, draft links are returned and user text is composed above quoted thread content.

* **Bug Fixes**
  * Improved plain-text handling by escaping special characters and preserving line breaks, with graceful fallback when compose defaults or signature data can’t be loaded.

### DIFF
--- a/plugins/outlook/src/compose-defaults.ts
+++ b/plugins/outlook/src/compose-defaults.ts
@@ -1,0 +1,241 @@
+import { owsRequest } from './outlook-api.js';
+
+/**
+ * Applies the user's Outlook compose defaults — default font and signature — to
+ * bodies the plugin creates, so plugin-authored drafts match what OWA's compose
+ * surface produces. These are client-side OWA behaviors that Graph's mailboxSettings
+ * does not expose; they are read from OWS gateway endpoints on the OWA origin.
+ */
+
+/**
+ * OWA's default compose font family. There is no server-stored compose font *name* —
+ * startup data carries only size/color/flags — so the family is OWA's client default
+ * (the Aptos stack), matching what the compose surface renders.
+ */
+const DEFAULT_FONT_FAMILY = 'Aptos, Aptos_EmbeddedFont, Aptos_MSFontService, Calibri, Helvetica, sans-serif';
+const DEFAULT_FONT_SIZE_PT = 12;
+const DEFAULT_FONT_COLOR = '#000000';
+
+/** 3- or 6-digit hex color, so a server value can't break out of the style attribute. */
+const HEX_COLOR_PATTERN = /^#[0-9a-fA-F]{3}(?:[0-9a-fA-F]{3})?$/;
+
+/**
+ * OWA's rich-text compose font-size enum → point size — the classic Outlook/Word
+ * font-size scale (1–7). `3 ⇒ 12 pt` is confirmed against the live compose surface;
+ * values outside this range fall back to the 12 pt default.
+ */
+const FONT_SIZE_PT_BY_ENUM: Record<number, number> = {
+  1: 8,
+  2: 10,
+  3: 12,
+  4: 14,
+  5: 18,
+  6: 24,
+  7: 36,
+};
+
+/** OWA compose font style bit flags (Microsoft rich-text convention). */
+const FONT_FLAG_BOLD = 1;
+const FONT_FLAG_ITALIC = 2;
+const FONT_FLAG_UNDERLINE = 4;
+
+/** Which signature a body should carry. */
+export type SignatureKind = 'new' | 'reply' | 'none';
+
+interface ComposeDefaults {
+  fontFamily: string;
+  fontSizePt: number;
+  fontColor: string;
+  fontFlags: number;
+  /** Display name of the signature applied to new messages, or null if none set. */
+  newSignatureName: string | null;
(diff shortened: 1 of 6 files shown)


Brandon-Haney:
Thanks for the review — addressed all four in the latest push (`a3ef020d`):

1. **`ComposeFontColor` validation** — the value is now checked against a hex-color pattern before it's interpolated into the `style` attribute, and falls back to the default otherwise.
2. **OWS origin** — switched from the hardcoded host to `window.location.origin` (read lazily, since the plugin module is also loaded in Node at build time where `window` is undefined) so the settings calls stay same-origin across `outlook.cloud.microsoft` / `outlook.office.com` / `outlook.office365.com`, and changed `credentials` to `same-origin`.
3 & 4. **Reply/forward draft cleanup** — moved `composeBody` + the PATCH inside the cleanup guard, so any pre-send failure (including in draft mode) deletes the created draft rather than leaving an orphan. A failure of the send itself still intentionally leaves the draft, since that outcome is ambiguous — the message may already be on its way, so deleting by draft id could remove a message that was actually sent.

The failing `e2e` check looks unrelated: it's the `side-panel-npm-search` letter-avatar-fallback test, which depends on the npm registry and flaked across all three retries (840/841 passed; `windows-e2e` is green).

Brandon-Haney:
Second round addressed in `7c88b53d`:

1. **`getComposeDefaults` JSDoc** — corrected. It now states the font lookup degrades to defaults while the signature-name lookup propagates whatever `owsRequest` throws (auth, rate-limit, non-OK, network).
2. **Shared compose-body assembly** — extracted `composeToolBody()` in `compose-defaults.ts` and used it from all four tools, so the `body_type` / `include_signature` → compose-options mapping lives in one place.

Two I'm leaving as-is, with reasoning:

3. **Full create → compose → cleanup → send lifecycle helper** — declining for now. The reply and forward handlers share the skeleton but differ materially (reply has `reply_all` + `createReply`/`createReplyAll`; forward has recipients + a `comment` rather than a body). Abstracting the whole lifecycle across two call sites with those differences adds more indirection than it removes at this scale, and `composeToolBody` already removes the substantive duplication.

4. **`encodeURIComponent(message_id)`** — declining in this PR, for consistency. The entire Outlook plugin passes raw message IDs into `api()` paths (get/update/move/delete/attachments as well as reply/forward); Graph returns URL-safe base64url IDs and the `=` padding is path-legal, which is why it works today. Encoding only these two sites would be inconsistent with the rest of the plugin without addressing the pattern globally — that's better as a separate plugin-wide change if wanted.

PR #138, opened 2026-07-09
feat(outlook): file attachments for compose tools (embedded + OneDrive link)

Brandon-Haney:
## What

Adds outgoing file attachments to the Outlook plugin's compose tools. Each attachment is either **embedded** as a copy of its bytes (`fileAttachment`) or, with `as_cloud_link`, uploaded to the user's OneDrive and attached as an organization-scoped sharing link (`referenceAttachment`).

Available on `create_draft`, `send_message`, `reply_to_message`, and `forward_message` via an `attachments[]` param, plus a standalone **`add_attachment`** tool for existing drafts.

| Type | Mechanism | Size |
| --- | --- | --- |
| Embedded, small | single inline `POST .../attachments` | ≤ 3 MB |
| Embedded, large | chunked Graph upload session | > 3 MB |
| Cloud link | OneDrive `PUT` + `createLink` → `referenceAttachment` | ≤ 250 MB |

## Design notes

- **Base64 input.** The plugin runs in the browser page with no filesystem, so a file's bytes arrive as a base64 string in the tool params (a `data:` prefix is stripped) — the same convention as the existing `microsoft-word_upload_file` / `wrike_upload_attachment` tools.
- **Attach on the draft's API base.** A message id minted on Graph is not valid on Outlook REST and vice versa, so attachments reuse the cached `mail` token that created the draft, and the attachment body is built per base (`#microsoft.graph.fileAttachment` vs `#Microsoft.OutlookServices.FileAttachment`; enum casing likewise).
- **Cloud uploads use a separate cache slot.** Drive writes need `Files` scope, which mail tokens lack, so `uploadAttachmentToOneDrive` cascades in its own slot restricted to Graph candidates.
- **`send_message` with attachments** switches from `sendMail` (which has no message id to attach to) to create-draft → attach → send; documented that the sent copy is always filed in Sent Items in that case.
- Refactors the auth-cascade boilerplate that `api()` and `owsRequest()` each hand-rolled into a single `cascadeAuth()` helper the new transports reuse.

## Testing

Verified live against a real enterprise mailbox and cleaned up afterward:

- `create_draft` + embedded file, and the standalone `add_attachment` (incl. `data:` prefix) — both land as `fileAttachment`.
- `create_draft` + `as_cloud_link` — OWA renders the cloud-link attachment (verified in the draft body).
- `reply_to_message` (draft) + embedded — attachment coexists with the reply signature and quoted thread.
- Chunked upload session — exercised end to end (`createUploadSession` → `Content-Range` chunk PUT → commit) by temporarily lowering the inline threshold so a small file routed through it; lands as an embedded `fileAttachment`.

Plugin `build` + `type-check` + `lint` + `format` all pass.

Builds on the compose-parity work in #137 (threaded drafts, signature, default font).

## Summary by CodeRabbit

* **New Features**
  * Outlook drafts, replies, forwards, and sent messages now support file attachments.
  * Attachments can be added inline or uploaded as cloud links for larger files.
  * Attachment support is now available as a standalone action for existing drafts.

* **Bug Fixes**
  * Improved handling of large attachments and upload errors.
  * Drafts are cleaned up automatically if attachment processing fails, helping avoid incomplete messages.

### DIFF
--- a/plugins/outlook/src/attachments.ts
+++ b/plugins/outlook/src/attachments.ts
@@ -1,0 +1,114 @@
+import { ToolError } from '@opentabs-dev/plugin-sdk';
+import { z } from 'zod';
+import {
+  attachFileToMessage,
+  attachLargeFileToMessage,
+  attachReferenceToMessage,
+  uploadAttachmentToOneDrive,
+} from './outlook-api.js';
+
+/**
+ * Adds attachments to draft messages the plugin creates. A file's bytes arrive as a
+ * base64 string in the tool params — the plugin runs in the browser page context and
+ * has no filesystem access, so a path on disk is not reachable. Each attachment is
+ * either embedded as a copy of its bytes (a `fileAttachment`) or, when
+ * `as_cloud_link` is set, uploaded to the user's OneDrive and attached as a sharing
+ * link (a `referenceAttachment`).
+ */
+
+/**
+ * Largest file embedded inline as base64 `contentBytes`. Microsoft rejects a single
+ * attachment request over ~3 MB, so anything larger needs a chunked upload session.
+ */
+const INLINE_ATTACHMENT_LIMIT_BYTES = 3_000_000;
+
+/** Shared input shape for one attachment, reused across every compose tool. */
+export const attachmentInputSchema = z.object({
+  name: z.string().min(1).describe('File name including extension, e.g. "report.pdf"'),
+  content_base64: z
+    .string()
+    .min(1)
+    .describe(
+      'File content, base64-encoded. A data: URI prefix (e.g. "data:application/pdf;base64,") is accepted and stripped.',
+    ),
+  content_type: z.string().optional().describe('MIME type (default: application/octet-stream)'),
+  as_cloud_link: z
+    .boolean()
+    .optional()
+    .describe(
+      'Attach as a OneDrive sharing link instead of embedding a copy of the file (default: false). Use for large files or documents recipients should open in the cloud.',
+    ),
+});
+
+export type AttachmentInput = z.infer<typeof attachmentInputSchema>;
+
+/** Strip an optional `data:` URI prefix, returning the bare base64 payload. */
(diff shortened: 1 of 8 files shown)


Brandon-Haney:
Both CodeRabbit findings addressed in the latest push:

1. **Large-attachment upload bound to Graph** — `attachLargeFileToMessage` now skips Outlook REST candidates (`apiBase !== GRAPH_API_BASE`) before calling `createUploadSession`, matching `uploadAttachmentToOneDrive`. Outlook REST v2.0 is decommissioned, so a non-Graph token would have thrown and aborted the cascade rather than falling through.

2. **`create_draft` attach made atomic** — it now guards the draft id and deletes the draft if the attach fails, matching `send_message`/`reply`/`forward`. No more orphaned draft on a partial-attach failure.

Brandon-Haney:
@opentabs-dev — nudging this before the stale bot closes it. I believe it's ready; no rush on your side.

Status since your last look:

- **CodeRabbit's two findings are addressed.** `attachLargeFileToMessage` now skips non-Graph candidates before opening the upload session, and `create_draft` deletes the draft if an attach fails so a partial failure can't orphan one — matching what `send_message`/`reply`/`forward` already did.
- **Verified live** against a real mailbox and cleaned up afterward: embedded attach, standalone `add_attachment` (incl. a `data:` prefix), cloud-link rendering in OWA, reply-with-attachment coexisting with the signature and quoted thread, and the chunked upload session end to end.
- **The one red check is unrelated.** `e2e › side panel npm search › search results without icons render letter-avatar fallback` failed on retry 2; 840 tests passed. That test queries the live npm registry and exercises the side panel, and this PR only touches `plugins/outlook/`. `main` has been green since, so a re-run should clear it.
- **Branch is 7 commits behind `main`, no conflicts** — none of those commits touch the outlook plugin. Happy to rebase if you'd rather see a fresh CI run against current `main`.

Whenever you have time.

PR #149, opened 2026-07-28
feat(teams): render message text as Markdown with native formatting

Brandon-Haney:
## What

`send_message` and `edit_message` now render their `text` input as **Markdown**, producing the native HTML dialect the Teams web client itself emits — fixing messages that previously lost line breaks and rendered Markdown lists as literal "1. 2. 3." text.

## Why

Both tools sent the raw `text` straight into a `RichText/Html` field with no conversion, so:

- plain-text newlines collapsed (HTML ignores insignificant whitespace), and
- `1. …` / `- …` lists rendered as run-on literal text instead of native lists.

## What's supported

The exact Teams HTML dialect was decoded from messages composed in the native editor, then a Markdown → Teams-HTML converter built to match it:

- **Structure:** headings (`#`–`###` → `<h1>`–`<h3>`), bulleted/numbered lists, GFM pipe tables, block quotes, horizontal rules (`---`), fenced code, paragraphs, soft line breaks
- **Inline:** bold, italic, strikethrough, inline code, links
- **Rich formatting** (via sanitized inline HTML, since Markdown cannot express these): underline, text color, highlight, and font size — matching what the composer toolbar offers. Color names map to Teams' fixed swatch palette (the exact hexes Teams stores).

## Safety

Inline HTML passes through a sanitizer: only a whitelist of tags is allowed, `style` is restricted to `color` / `background-color` / `font-size` with palette-validated values, `href` schemes are limited to `http`/`https`/`mailto`, and everything else is HTML-escaped — so a message never carries markup Teams won't render or an injection vector.

## Verification

Verified live against the Teams web app: messages were sent through the tools and read back, confirming the stored HTML matches Teams' native dialect exactly (including the 8 text-color and 8 highlight swatch hexes, the three font sizes, and `<hr>`), and confirming visual rendering in the client. Off-palette colors and unknown tags were correctly dropped/escaped.

## Summary by CodeRabbit

* **New Features**
  * Added Markdown-to-Microsoft Teams rich-text conversion for message content.
  * Messages now render common Markdown features (headings, lists, tables, quotes, code blocks, links, emphasis) and a limited set of inline HTML.
  * Markdown formatting is applied consistently for both sending new messages and editing existing ones.
  * Updated tool input and descriptions to clearly indicate Markdown support and formatting guidelines.

### DIFF
--- a/plugins/teams/src/markdown-html.ts
+++ b/plugins/teams/src/markdown-html.ts
@@ -1,0 +1,432 @@
+// Teams chat messages are sent with `messagetype: 'RichText/Html'`, so the
+// `content` field is parsed as an HTML fragment. This module converts a
+// message written in Markdown into the exact HTML dialect the Teams web client
+// itself emits (decoded from messages composed in the native editor), and lets
+// a curated set of inline HTML tags pass through for the formatting Markdown
+// cannot express — text colour, highlight, font size, and underline — matching
+// what a human can apply from the composer toolbar.
+//
+// Markdown -> Teams HTML:
+//   **bold**      -> <strong>            *italic* / _italic_ -> <i>
+//   ~~strike~~    -> <s>                 `code`              -> <code>
+//   [text](url)   -> <a href title>      ```fence```         -> <pre><code>
+//   - item        -> <ul><li>            1. item             -> <ol><li>
+//   # heading     -> <h1>…<h3>           > quote             -> <blockquote>
+//   | a | b |     -> <table>             ---                 -> <hr>
+//   paragraph     -> <p>…</p>            soft line break     -> <br>
+//
+// Inline HTML passthrough (sanitised — every other tag/attribute is escaped):
+//   <u>…</u>                                  underline
+//   <span style="color:NAME">…</span>         text colour  (NAME is a swatch)
+//   <span style="background-color:NAME">…</span> highlight  (NAME is a swatch)
+//   <span style="font-size:large|medium|small">…</span> font size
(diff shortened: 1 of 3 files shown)


Brandon-Haney:
Thanks for the review — all actionable items addressed in ef80227e:

- **Critical · Markdown link scheme bypass** — extracted a shared `safeHref` helper and applied it to both the `[label](url)` path and the inline `<a>` path, so non-`http(s)`/`mailto` schemes (e.g. `javascript:`) drop the href on both.
- **Major · prototype-chain lookups** — guarded the palette and font-size lookups with `Object.hasOwn`, so keys like `constructor` no longer resolve through `Object.prototype` into the emitted style.
- **Correctness · intra-word `_`** — `_`/`__` now toggle emphasis only at word boundaries (GFM intraword rule), so `snake_case` and `a_b_c` stay literal; `*`/`**` are unchanged.
- **Doc comment** — corrected `parseInline` to describe the toggle semantics (an unmatched delimiter stays active to end of line, not literal).
- **DRY** — the duplicated tool-description blurb is now a single exported `MARKDOWN_FORMATTING_HELP` constant used by both tools.

**Deferred · unit tests** — the converter is pure and very testable, but this plugin (like the others in the repo) has no unit-test runner configured, so standing one up is a repo-wide decision rather than part of this PR. The converter was verified live against Teams (send → read-back → visual render) plus targeted checks for each fix above.

Brandon-Haney:
@opentabs-dev is this project still active?

